PR #39, opened 2022-04-12
messagepack unpacker set use_list argument to False by default

AlirezaSohofi:
Set the default parameter of the use_list for msgpack.Unpacker to False, as suggested by the authors (see performance tips [here](https://pypi.org/project/msgpack/))

# Description

Fixes # issue

## Type of change

- [ ] Documentation update
- [x] Other (please describe): set the default value of use_list for msgpack unpacker for better performance.

# Checklist:


squirrel/serialization/msgpack.py
```diff
@@ -56,7 +56,11 @@ def serialize_shard_to_file(
 
     @staticmethod
     def deserialize_shard_from_file(
-        fp: str, fs: t.Optional[AbstractFileSystem] = None, mode: str = "rb", **open_kwargs
+        fp: str,
+        fs: t.Optional[AbstractFileSystem] = None,
+        mode: str = "rb",
+        unpacker_kwarg: t.Optional[t.Dict] = None,
```
AlpAribal, line 62: How about an `s` here
```suggestion
        unpacker_kwargs: t.Optional[t.Dict] = None,
```

squirrel/serialization/msgpack.py
```diff
@@ -73,9 +79,12 @@ def deserialize_shard_from_file(
...
+        unpacker_kwarg = {} if unpacker_kwarg is None else unpacker_kwarg
+        if "use_list" not in unpacker_kwarg:
+            unpacker_kwarg["use_list"] = False
 
         if fs is None:
             fs = fsspec
 
         with fs.open(fp, **open_kwargs) as f:
-            yield from msgpack.Unpacker(f, object_hook=msgpack_numpy.decode)
+            yield from msgpack.Unpacker(f, object_hook=msgpack_numpy.decode, **unpacker_kwarg)
```
AlpAribal, line 90: I guess object hook can also be provided via `unpacker_kwarg`, we can as well check for it at the start and make a note of it in arg description

squirrel/serialization/msgpack.py
```diff
@@ -65,6 +69,8 @@ def deserialize_shard_from_file(
             fs (AbstractFileSystem, optional): Filesystem to use for opening the file. If not provided, `fsspec` will
                 pick a filesystem suitable for `fp`. Defaults to None.
             mode (str): IO mode to use. Passed to :py:meth:`fs.open`. Defaults to "rb".
+            unpacker_kwarg (Dict, optional): kwargs to be passed to `msgpack.Unpacker()`.
```
AlpAribal, line 72: To match the case of other args.
```suggestion
            unpacker_kwarg (Dict, optional): Kwargs to be passed to `msgpack.Unpacker()`.
```

AlpAribal (changes requested):
Left minor comments.

squirrel/serialization/msgpack.py
AlpAribal: What I mean is that currently, when the user passes the `"object_hook"` kwarg as part of `unpacker_kwargs`, it will fail since the kwarg is passed twice to `Unpacker()`. I suggest we move `object_hook` into `unpacker_kwargs` as well and then just pass `**unpacker_kwargs` to `Unpacker()`

AlpAribal (approved):
LGTM!

PR #44, opened 2022-05-05
accept callable as source for composable to make it completly lazy

AlirezaSohofi:
# Description

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:


squirrel/driver/store_driver.py
```diff
@@ -77,4 +82,6 @@ def keys(self, **kwargs) -> Iterable:
     @property
     def store(self) -> AbstractStore:
         """Store that is used by the driver."""
+        if self._store is None:
+            self._store = SquirrelStore(url=self.url, serializer=self.serializer)
```
AlpAribal, line 86: With this, StoreDriver can only use `SquirrelStore` as its underlying store. Maybe we should also have a setter so that "technically" it still works with `AbstractStore`

PR #47, opened 2022-05-20
introduce loop and fixed size iterable

AlirezaSohofi:
This pr introduces two changes:
- `loop()` method to iterate over the IterableSource `n` time
- `take()` is updated, so that it loops over the iterable until exactly `n` items are returned 

# Description

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:


.gitignore
```diff
@@ -120,6 +120,7 @@ target/
 
 # IPython Notebook
 .ipynb_checkpoints
+Untitled*.ipynb
```
AlpAribal, line 123: Is this change intended?

squirrel/iterstream/iterators.py
```diff
@@ -72,6 +73,13 @@ def shuffle_(
         yield _pick(buf, rng)
 
 
+def loop_(iterable: t.Iterable, n: int) -> t.Iterator:
+    """Loop over the iterable n times"""
+    iterable = itertools.tee(iterable, n)
```
AlpAribal, line 78: `itertools.tee` [docs](https://docs.python.org/3/library/itertools.html#itertools.tee) state that `"In general, if one iterator uses most or all of the data before another iterator starts, it is faster to use list() instead of tee()"`. I think this is that case. AFAIU, `tee()` will keep all elements in the iterable in memory anyway. We can as well iterate through the iterable while saving elements in a deque. Then we can iterate through that deque for repeats > 1. This way, we can have infinite loops too

squirrel/iterstream/base.py
```diff
@@ -79,6 +81,20 @@ def take(self, n: t.Optional[int]) -> Composable:
             return self
         return self.to(take_, n)
 
+    def take_exact(self, n: int) -> Composable:
```
AlpAribal, line 84: If what I suggest for `loop()` makes sense to you, `take_exact()` is not necessary, one can do e.g. `it.loop(n=None).take(N)`

AlpAribal (changes requested):
What do you think about keeping the deque ourselves instead of using `tee()`?

squirrel/iterstream/base.py
```diff
@@ -79,6 +80,20 @@ def take(self, n: t.Optional[int]) -> Composable:
...
+        """Yield the first n elements from the iterable. Less elements can be yielded if the iterable does not have
+        enough elements. If n is bigger than the number of items in the iterable, this method will loop until
+        *exactly* n items are yielded.
+
+        Args:
+            n (int): Number of samples to take.
+        """
+        return _FixedLengthIterable(self, n)
+
+    def loop(self, n: int) -> Composable:
```
AlpAribal, line 93: How about `repeat` instead of `loop`?

squirrel/iterstream/base.py
```diff
@@ -79,6 +80,20 @@ def take(self, n: t.Optional[int]) -> Composable:
...
+        """Repeat the iterable n times"""
```
AlpAribal, line 94: I would be for allowing `n=None` to have infinite loops. Then, we do not need to implement `take_exact()`, one can do `loop(None).take(N)`.

AlpAribal (changes requested):
`loop()` currently does not work with sources that are generators, which is important to support. Here is an example:
```
from typing import Iterable

my_generator = (i for i in range(5))
print(isinstance(my_generator, Iterable))  # True
it = IterableSource(my_generator)
res = it.loop(2).collect()  # raises cannot pickle 'generator' object in deepcopy() call
```
How about using `tee` again or having our own implementation that uses a deque?

squirrel/iterstream/base.py
```diff
@@ -79,6 +80,20 @@ def take(self, n: t.Optional[int]) -> Composable:
             return self
         return self.to(take_, n)
 
+    def take_exact(self, n: int) -> Composable:
+        """Yield the first n elements from the iterable. Less elements can be yielded if the iterable does not have
```
winfried-loetzsch, line 84: The current implementation of take_exact also doesn't work with generators, for the same reason Alp mentioned above.

winfried-loetzsch (changes requested):
I agree with Alps feedback from before - especially that `take_exact` is a bit counter intuitive for me at the first read and I would find something like `repeat(None).take(x)` more intuitive. Besides that looks good!

AlirezaSohofi:
@AlpAribal @winfried-loetzsch Thanks for the feedback. I think `loop(n=None).take()` is a good suggestion, will implement it this way. `repeat` might be confusing, as it's less clear (in my opinion) if it repeats elements or the iterable. 

Regarding the generator issue: it's not an isolated issue, every time you have a non-picklable object (e.g. fsspec, generator, etc) in your pipeline you face this problem. The solution is to not instantiate non-picklable objects in the `__init__`, but in the `__iter__`. This is currently happening in `get_iter()` for instance, `SquirrelStore.keys()` which is needed by `get_iter()`, returns a generator. we simply avoid calling this method until the `__iter__` method is called. Let me re-write alp's example this way:

```
def gen():
    yield from range(5)

class MyLazyGenerator(Composable):
    def __init__(self, g):
        self.g = g
    
    def __iter__(self):
        yield from self.g()

it = MyLazyGenerator(gen).loop(2).collect()
```
This is admittedly not well explained in the documentation, @pzdkn is putting a lot of thoughts into it atm.  

@AlpAribal we can't keep data in memory, `tee` and deque would require that, please correct me if I am wrong. If you have any suggestion that avoids keeping the data in memory AND `deepcopy`, that would be awesome!

test/test_iterstream/test_stream.py
```diff
@@ -58,6 +58,33 @@ def test_take(samples: t.List[SampleType]) -> None:
     assert len(res) == len(samples)
 
 
+def test_take_exact(samples: t.List[SampleType], num_samples: int) -> None:
+    """Test take_exact"""
+
+    # take all elements in iterator
+    it = IterableSource(samples).loop().take(len(samples)).collect()
+    assert len(list(it)) == len(samples)
```
AlpAribal, line 66: Do we need the list-casting here (since we are collecting the iterable in the line above)? If no, we can remove it. There are a few other ones below as well
```suggestion
    assert len(it) == len(samples)
```

test/test_iterstream/test_stream.py
```diff
@@ -58,6 +58,33 @@ def test_take(samples: t.List[SampleType]) -> None:
...
+    assert len(list(it)) == 0
+
+    # ensure that empty iterable works as expected
+    it = IterableSource([]).loop().take(2).collect()
+    assert len(list(it)) == 0
+
+
+@pytest.mark.parametrize("n", [0, 2, 4, 8, 10])
+def test_loop(samples: t.List[SampleType], n: int) -> None:
+    """Test loop"""
```
AlpAribal, line 83: We should also test that each loop has the same order of elements

squirrel/iterstream/base.py
```diff
@@ -242,6 +254,55 @@ def __iter__(self) -> t.Iterator:
         return self.f(iter(self.source), *self.args, **self.kw)
 
 
+class _FixedLengthIterable(Composable):
```
AlpAribal, line 257: We don't need this class anymore, right?

squirrel/iterstream/base.py
```diff
@@ -16,8 +17,8 @@
     map_,
     monitor_,
     shuffle_,
-    take_,
     tqdm_,
+    take_,
```
AlpAribal, line 21: The old order makes more sense to me

squirrel/iterstream/base.py
```diff
@@ -242,6 +254,55 @@ def __iter__(self) -> t.Iterator:
...
+                self._started = True
+                self.idx += 1
+            except StopIteration:
+                if not self._started:
+                    return
+                current_ = iter(deepcopy(self.source))
+
+
+class _LoopIterable(Composable):
+    def __init__(self, source: t.Iterable, n: int):
```
AlpAribal, line 283:
```suggestion
    def __init__(self, source: t.Iterable, n: Optional[int]):
```

test/test_iterstream/test_stream.py
```diff
@@ -58,6 +58,33 @@ def test_take(samples: t.List[SampleType]) -> None:
     assert len(res) == len(samples)
 
 
+def test_take_exact(samples: t.List[SampleType], num_samples: int) -> None:
```
AlpAribal, line 61:
```suggestion
def test_take_exact(samples: t.List[SampleType]) -> None:
```

AlpAribal (dismissed):
LGTM! Left minor comments, feel free to merge after addressing and bumping the version.

test/test_iterstream/test_stream.py
```diff
@@ -90,6 +90,33 @@ def test_take(samples: t.List[SampleType]) -> None:
...
+    it = IterableSource(samples).loop().take(0).collect()
+    assert len(it) == 0
+
+    # ensure that empty iterable works as expected
+    it = IterableSource([]).loop().take(2).collect()
+    assert len(it) == 0
+
+
+@pytest.mark.parametrize("n", [0, 2, 4])
+def test_loop(samples: t.List[SampleType], n: int) -> None:
```
AlpAribal, line 114: We should also test the order of elements, i.e. `IterableSource([1,2,3]).loop(3).colllect() = [1,2,3,1,2,3,1,2,3]`

test/test_iterstream/test_stream.py
```diff
@@ -98,11 +125,6 @@ def test_take_side_effect() -> None:
     assert list(take_(it, 2)) == [3, 4]
 
 
-def test_take_less_elements() -> None:
-    """Check that trying to take more elements than possible does not lead to errors."""
-    assert list(take_([1, 2, 3], 10)) == [1, 2, 3]
```
AlpAribal, line 103: On second thought, the new tests do not cover this case. We can as well keep this one too.

AlpAribal (changes requested):

PR #49, opened 2022-06-03
CatalogKey can be used to index catalog

AlpAribal:
# Description

Makes it possible to get, set, and delete sources in a catalog using a `CatalogKey` or a `(id, version)` tuple.

## Type of change

- [ ] Documentation update

# Checklist:


squirrel/catalog/catalog.py
```diff
@@ -71,22 +72,54 @@ def __eq__(self, other: Any) -> bool:  # noqa D105
...
         if identifier not in self:
-            # return a dummy object to let the user set a version directly
-            return DummyCatalogSource(identifier, self)
-        return self.sources[identifier][-1]
+            if isinstance(identifier, str):
+                # return a dummy object to let the user set a version directly
+                return DummyCatalogSource(identifier, self)
+            else:
+                # DummyCatalogSource only allows setting the version after initialization, cannot set it at this point
+                raise KeyError
```
AlirezaSohofi, line 118: please give an informative error message

AlirezaSohofi: would be cool to have the test case for this too

squirrel/catalog/catalog.py
```diff
@@ -71,22 +72,54 @@ def __eq__(self, other: Any) -> bool:  # noqa D105
...
+            version = -1
+        else:
+            identifier, version = identifier
 
-    def __getitem__(self, identifier: str) -> CatalogSource:  # noqa D105
         if identifier not in self:
-            # return a dummy object to let the user set a version directly
-            return DummyCatalogSource(identifier, self)
-        return self.sources[identifier][-1]
+            if isinstance(identifier, str):
```
AlirezaSohofi, line 113: why do we need this additional check `if isinstance(identifier, str):` ?

squirrel/catalog/catalog.py
```diff
@@ -71,22 +72,54 @@ def __eq__(self, other: Any) -> bool:  # noqa D105
...
-    def __delitem__(self, identifier: Union[str, CatalogKey]) -> None:  # noqa D105
-        del self._sources[identifier]
+    def __contains__(self, identifier: Union[str, CatalogKey, Tuple[str, int]]) -> bool:  # noqa D105
+        if isinstance(identifier, str):
+            return identifier in self._sources
+
+        identifier, version = identifier
+        return identifier in self._sources and version in self._sources[identifier]
+
+    def __delitem__(self, identifier: Union[str, CatalogKey, Tuple[str, int]]) -> None:  # noqa D105
```
AlirezaSohofi, line 82: the behaviour might be unexpected, you can overwrite an existing version, or set and arbitrary version, etc, should we explain it some where?

test/test_catalog/test_catalog.py
```diff
@@ -22,16 +23,25 @@ def test_catalog_createsimple() -> None:
 
 
 def test_catalog_versioning() -> None:
-    """Test versioning source in a Catalog"""
+    """Test versioning a source in a Catalog using all allowed identifier types."""
```
AlirezaSohofi, line 26: please add test cases for `__delitem__`, `__contains__`, `__getitem__` for non-existing CatalogKey, and providing unsupported CatalogKey (e.g. non-string identifier?)

AlirezaSohofi (changes requested):

squirrel/catalog/catalog.py
AlpAribal: We allow two types of identifiers:
1) `str`: Just the id of the catalog source, the user gets the version they want later on, e.g. `cat["my_id"][my_ver]`
2) `Tuple[str, int]`: This is the new format, basically either a `CatalogKey` or a tuple similar to it. This is what this PR introduces. The user can specify the version directly, e.g. `cat["my_id", my_ver]`.

To differentiate btw the two types, checking if identifier is a string is easier. If we check `isinstance(identifier, CatalogKey)`, then we do not cover the `Tuple[...]` case, and if we check `isinstance(identifier, Tuple)` then we do not cover the `CatalogKey` case. So, we check if identifier is string and if not we conclude that the identifier is tuple-like.

squirrel/catalog/catalog.py
AlpAribal: Which behavior exactly? The string identifier case is same as before. Do you mean we should remove all version also when a tuple identifier is provided?

AlirezaSohofi (dismissed):
LG :+1:

PR #56, opened 2022-06-22
zip multiple iterables as a source

AlirezaSohofi:
# Description

The use case: we have a store with samples, and `1..n` other stores that each contain only features. These stores must have the same keys and same number of samples per shard.  

`IterableZipSource` makes it possible to zip items from several iterables and use that as a source. For instance:

```
it1 = MessagepackDriver(url1).get_iter()
it2 = MessagepackDriver(url2).get_iter()

it3 = IterableZipSource(iterables=[it1, it2]).collect()

```

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/squirrel/iterstream/source.py
+++ b/squirrel/iterstream/source.py
@@ -139,3 +139,12 @@ class IterableSamplerSource(Composable):
                     self.probs = [p / s for p in probs]
             if len(self.iterators) == 0:
                 break
+
+
+class IterableZipSource(Composable):
+
+    def __init__(self, iterables: t.List[t.Iterable]):
+        super().__init__(source=iterables)
+
+    def __iter__(self):
+        yield from zip(*self.source)
--- a/squirrel/iterstream/source.py
+++ b/squirrel/iterstream/source.py
@@ -142,6 +142,7 @@ class IterableSamplerSource(Composable):
 
 
 class IterableZipSource(Composable):
+    """Makes it possible to zip items from several iterables and use it as a source."""
 
     def __init__(self, iterables: t.List[t.Iterable]):
         super().__init__(source=iterables)


mg515 (approved):
Thanks, LG!

AlirezaSohofi:
@AlpAribal 

what do you think about `ZipDriver` instead? It would accept an iterable of drivers, which should all have the exact same keys and number of items in each shard, and produces a tuple of samples retrieved from each driver.

AlpAribal:
> what do you think about `ZipDriver` instead? It would accept an iterable of drivers, which should all have the exact same keys and number of items in each shard, and produces a tuple of samples retrieved from each driver.

I am not sure how common of a use case this is, so it requires a separate driver. If we see that separating sample features into different stores is unavoidable/meaningful, and converge to such storage patterns, then a zipping driver would definitely make sense.
Though, I feel it will get harder to maintain the more shallow wrappers we have around zip/itertools/more_itertools.

AlirezaSohofi:
> If we see that separating sample features into different stores is unavoidable/meaningful, and converge to such storage patterns, then a zipping driver would definitely make sense.

totally agree. Additionally, the performance would be worse, but it would be more flexible. 

> I am not sure how common of a use case this is

less figure this out.

mg515:
> I am not sure how common of a use case this is

I think it would be a fairly common case. Imagine a ML pipeline starting with writing raw or transformed data in a store. Later, you compute some features for each of the sample. You need another source for that, unless you write in the existing store (I don't think so).
Even later on you might want to compute predictions for each of the models you developed on the same sample set. You generate more and more elements, relating to the sample. It would be cool if there was a way to reliably zip those together in a streaming fashion to e.g. make an analysis.

pzdkn:
I think this PR is useful. I think a iterstream source can be helpful for mapping functions over multiple iterstreams.  This can be helpful independant of drivers.

AlirezaSohofi:
FYI @MaxSchambach

pzdkn:
Actually the new API looks quite good, thanks @AlirezaSohofi

PR #64, opened 2022-06-28
Store the processing steps in a stream

pzdkn:
# Description

Store more information in `Composables`:

- Which Squirrel version is used
- Git info e.g. commit-hash, remote repository
- Log processing steps when chaining `Composables`

This aims to provide the user more information about the stream. When a `Composable` stores sensitive information e.g. url in `FilePathGenerator`, then this should not be logged.

Fixes # issue

## Type of change

- [ ] Documentation update
- [x] Refactoring including code style reformatting 

# Checklist:


test/test_iterstream/test_stream.py
```diff
@@ -20,6 +20,21 @@
...
+    """Test if history is correctly appended when chaining Composables"""
+
+    def add_1(x):
+        return x + 1
+
+    def mult_10(x):
+        return x * 10
+
+    it = IterableSource(samples).map(add_1).map(mult_10)
+    assert len(it.history) == 2
```
AlirezaSohofi, line 33: why not 3?

test/test_iterstream/test_stream.py
```diff
@@ -20,6 +20,21 @@
...
+    assert it.history[0]['args'][0] == 'add_1' and it.history[1]['args'][0] == 'mult_10'
```
AlirezaSohofi, line 34: suggestion: `add_1` --> `my_project.my_module.my_func`

squirrel/iterstream/base.py
```diff
@@ -31,9 +31,10 @@ class Composable:
     def __init__(self, source: t.Optional[t.Union[t.Iterable, t.Callable]] = None):
         """Init"""
         self.source = source
+        self.history = []
```
AlirezaSohofi, line 34: history should be private

squirrel/iterstream/base.py
```diff
@@ -209,12 +212,24 @@ def source_(self, source: Composable) -> Composable:
...
         """When items in the stream are themselves iterables, flatten turn them back to individual items again"""
         return self.to(flatten_)
 
+    def add_to_history(self, parent: Composable,
+                       function_or_class: t.Union[t.Callable, t.ClassVar],
+                       *args, **kwargs) -> None:
+        """Adds a processing step entry to the history"""
+        step = {'name': function_or_class.__name__,
+                'args': [arg.__name__ if callable(arg) else arg for arg in args],
+                'kwargs': kwargs}
```
AlirezaSohofi, line 229: would be cool if we can include some information about the git hash

AlirezaSohofi (commented):
Looks very good :1st_place_medal:

AlpAribal:
> The alternative approach is to loop until no `source` attribute exists anymore and log the processing steps in the loop. What do you prefer?

I would tend towards this lazy approach but do not feel strong enough to reject the current solution.

squirrel/iterstream/base.py
```diff
@@ -31,9 +33,11 @@ class Composable:
     def __init__(self, source: t.Optional[t.Union[t.Iterable, t.Callable]] = None):
         """Init"""
         self.source = source
+        self._history = []
```
AlirezaSohofi, line 36: please also add a getter property to retrieve the history.

AlirezaSohofi: Maybe `steps` is a more descriptive name than `history`.

squirrel/iterstream/base.py
```diff
@@ -31,9 +33,11 @@ class Composable:
...
+        self.add_to_history(self, self.__class__)
```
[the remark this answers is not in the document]
AlirezaSohofi: Source can be something other than `Composable`, e.g. the `IterableSource` wraps an iterable or a callable.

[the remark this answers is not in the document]
AlpAribal: Note that an `IterableSource` is also a `Composable`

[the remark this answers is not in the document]
AlirezaSohofi: Then I probably did not understood your point correctly, I thought you mean checking the `source` attribute of the `Composable` object is `Composable`, which is the case for intermediate steps of the pipeline but not necessarily for the first step (e.g. `IterableSource`).

squirrel/iterstream/base.py
pzdkn: steps is more fitting, true!

squirrel/iterstream/base.py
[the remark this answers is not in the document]
pzdkn: I think we still don't get around to call `add_to_hist()` in `.to()` and. `.compose()` as we also want to log the function and function parameters, right?

test/test_iterstream/test_stream.py
```diff
@@ -239,9 +276,9 @@ def test_metrics_tracking_with_mlflow(
         with mlflow.start_run(run_name="squirrel-test"):
             (
                 it.monitor(mlflow.log_metrics, prefix="before shuffle ", metrics_conf=conf)
-                .shuffle(20)
-                .monitor(mlflow.log_metrics, prefix="after shuffle ", metrics_conf=conf)
-                .collect()
+                    .shuffle(20)
+                    .monitor(mlflow.log_metrics, prefix="after shuffle ", metrics_conf=conf)
+                    .collect()
```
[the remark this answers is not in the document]
pzdkn: thanks for the hint 😄

squirrel/iterstream/base.py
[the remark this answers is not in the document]
AlpAribal: With the current solution, in `Composable.to()`, we have:
```python
it = _Iterable(self, f, *args, **kw)
it.add_to_history(self, f, *args, **kw)
```
So, we first create a new Composable and then add the previous Composable to its history. To me, having a design where the addition to history is separated from the instantiation of the Composable is unwanted since
1. I can forget to add the parent to history when I create a new method, say `Composable.my_new_method()`, in which I create a new Composable and return this newly created object (similar to `.to()`)
2. We kind of signal that it is ok/expected to manage the history from the outside, i.e. the Composable itself does not know how the history is managed, we have to manage it for it when we are creating it

Having the separation can lead to unexpected results. Consider the following case:
```python
def add_1(x):
    return x + 1

def mult_10(x):
    return x * 10

it = IterableSource([1, 2, 3]).map(add_1).map(mult_10)
it2 = IterableSource(it)

len(it._history)  # 3
len(it2._history)  # 1
```
Here, for `it2`, I would expect a history of length 4, now we get 1 since `it2` is not aware that there actually is some history before itself and no one is managing the history for it.

What I suggest is that Composable checks whether its source is another Composable (the parent). If this is the case, it sets its own history as the history of the parent + one more step.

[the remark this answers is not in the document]
pzdkn: What I don't quite understand is how you can retrieve the function and functional arguments from the constructor of `Composable` as it only receives `source`.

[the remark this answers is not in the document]
AlpAribal: We can define a `Compoasble._add_self_to_history()` method, which is called in `Composable.init()` to, ehm, add self to history. Then, the subclasses (`_Iterable`, `_AsyncMap`, `_LoopIterable`, ...) can override this method, if necessary, such as the case you point out with `_Iterable`.

squirrel/iterstream/base.py
```diff
@@ -31,6 +34,18 @@ class Composable:
     def __init__(self, source: t.Optional[t.Union[t.Iterable, t.Callable]] = None):
         """Init"""
         self.source = source
+        self._add_to_steps()
```
AlpAribal, line 37: How about calling this method in `steps` property? We can set `self._step = None` here and then
```python
    @property
    def steps(self) -> t.List[t.Dict[str, t.Union[str, t.Dict]]]:
        """Getter for steps"""
        if self._steps is None:
            self._add_to_steps()
        return self._steps
```
The reason I am suggesting this is that now the subclasses of `Composable` need to know that they should set their variables before calling `super().__init__()` in order for their variables to be shown in `steps`. I see that we have considered this in `_Iterable` but there are other subclasses as well: `_LoopIterable`, `_AsyncMap`, `_NumbaMap`.

Another plus is  that we won't be keeping unnecessary memory if the user does not care about the steps.

It would also be nice to have tests for checking the steps args for each Composable subclass.

pyproject.toml
```diff
@@ -6,4 +6,4 @@ recursive = true
 
 [tool.black]
 line-length = 120
-include = '\.pyi?$'
+include = '\.pyi?$'
```
AlpAribal, line 9: Is this change on purpose?

squirrel/iterstream/base.py
```diff
@@ -212,19 +222,40 @@ def numba_map(self, callback: t.Callable) -> _NumbaMap:
...
+                continue
+            if k == "args":
+                class_args[k] = [get_callable_name(obj) for obj in v]
+            else:
+                class_args[k] = get_callable_name(v)
+
+        step = {
+            "class": get_callable_name(self.__class__),
+            "class_args": class_args,
+            "git_version": subprocess.check_output(["git", "describe"]).strip().decode(),
```
AlpAribal, line 248: This fails if we are running squirrel outside a git repo. If we are to keep this, we should wrap this with a try/except.

squirrel/iterstream/base.py
pzdkn: yeah, quite a nice idea. Thanks :)

pyproject.toml
pzdkn: Semi, I tried to change my black config back to default. I will revert it back.

squirrel/iterstream/base.py
```diff
@@ -226,6 +230,43 @@ def monitor(
...
+        def get_callable_name(obj: t.Any) -> str:
+            """Obtain the name of a callable with the module it is in"""
+            return f"{inspect.getmodule(obj).__name__}.{obj.__name__}" if callable(obj) else obj
+
+        class_args = dict()
+        for k, v in self.__dict__.items():
+            if k in ["source", "_steps"]:
+                continue
+            if k == "args":
+                class_args[k] = [get_callable_name(obj) for obj in v]
```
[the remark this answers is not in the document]
AlpAribal: I am asking this to better understand the use case.
- If we want steps to have a neat string form / representation, then we can make `get_callable_name` recursive so that all iterables are further prettified
- If we want steps (and all args of the steps) to be programmatically accessible (so that I can maybe test a function arg in the pipeline when debugging, e.g. `it.steps[-1]["class_args"]["args"][0](my_test_input)`, then leaving functions as they are might be better
@pzdkn @AlirezaSohofi wdyt?

test/test_iterstream/test_stream.py
```diff
@@ -13,13 +13,44 @@
...
 
+def test_streamsteps() -> None:
+    """Test logging steps when chaining Composables using map"""
+
+    def add_1(x: float) -> float:
+        """Add 1 to x"""
+        return x + 1
+
+    def mult_10(x: int) -> float:
+        """Mult 10 to x"""
```
AlpAribal, line 32:
```suggestion
        """Mult x with 10"""
```

squirrel/iterstream/base.py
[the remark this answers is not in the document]
pzdkn: - It's true, it's targeting the `_Iterable` class.
- Not really sure what the recursive solution should look like. Do you mind elaborating @AlpAribal ?

AlirezaSohofi (changes requested):
There are some remaining issues:
1. `len(IterableSource([1,2,3]).compose(MyIter).steps) == 2` --> False. Please investigate why and add test for it
2. if there is a big object, say a list, in the arguments, it must be excluded. We can try to provide some information like length. `steps` should only contain small objects so that we can log it easily. An example: `IterableSamplerSource([[1,2,3], [11,22,33]]).steps`
3. `steps` should only contains basic data types, otherwise it might not be possible to log it. This must be checked for user provided arguments (similar to `2`). Example in `2` contains a `rng` object.
4.  `FilePathGenerator("/home/user/data").steps` gives you the url in clear text which could be a security issue. This should be somehow masked.

AlirezaSohofi:
> 4\. `FilePathGenerator("/home/user/data").steps` gives you the url in clear text which could be a security issue. This should be somehow masked.

It would be useful to let the user decide on this. For that, `steps` should be changed to `steps(secure=False)`

pzdkn:
> 3\. `steps` should only contains basic data types, otherwise it might not be possible to log it. This must be checked for user provided arguments (similar to `2`). Example in `2` contains a `rng` object.

So we would only store primitive data-types? For Iterables, we store the length?

PR #65, opened 2022-06-29
[DRAFT] Support for different SquirrelStore compression modes

### LINKED ISSUE #59: [FEATURE] Allow configuring compression mode in MessagepackSerializer
Hey,

Thank you for working on this library!
I think it has a huge potential, especially for dataset creators to provide their dataset in an optimized deep-learning format that is well suited for distribution.
The performance of the MessagepackSerializer is amazing and being able to distribute subsets of the dataset (shards) is something I never wanted but really want to utilize in the future!

I have played around with some "MessagepackSerializer" configurations and according to some internal benchmarks, it would be helpful to allow the user to configure the compression algorithm. 

https://github.com/merantix-momentum/squirrel-core/blob/0ff6be7368a258d02e7c39d63ed2993d8a5af322/squirrel/serialization/msgpack.py#L28-L48

Currently, the compression mode is "locked" to gzip.
I assume the main reason is due to the wide usage of gzip and to keep the code 'simple' as it makes it easy in the deserializer to _know_ that the gzip compression was used:

https://github.com/merantix-momentum/squirrel-core/blob/0ff6be7368a258d02e7c39d63ed2993d8a5af322/squirrel/serialization/msgpack.py#L58-L81

Here I would like to note that given the extension, `fsspec` (default) _could_ also _infer_ the compression by inspecting the [filename suffix](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.open). But I can see how this might cause problems if somebody would like to switch out `fsspec` with something else (although I would have no idea with what and why :D )

Other spots within the codebase that are _coupled_ to this compression assumption are the methods from the `SquirrelStore`:

https://github.com/merantix-momentum/squirrel-core/blob/0ff6be7368a258d02e7c39d63ed2993d8a5af322/squirrel/store/squirrel_store.py#L12-L67

Or to show the significant parts:
- get:
https://github.com/merantix-momentum/squirrel-core/blob/0ff6be7368a258d02e7c39d63ed2993d8a5af322/squirrel/store/squirrel_store.py#L40-L41

- set:
https://github.com/merantix-momentum/squirrel-core/blob/0ff6be7368a258d02e7c39d63ed2993d8a5af322/squirrel/store/squirrel_store.py#L59-L60

- keys:
https://github.com/merantix-momentum/squirrel-core/blob/0ff6be7368a258d02e7c39d63ed2993d8a5af322/squirrel/store/squirrel_store.py#L66-L67

In my internal benchmarks, I was able to greatly speed up the data loading by simply using no compression at all (`None`).
Although I am fully aware that the _correct_ compression mode heavily depends on the specific hardware/use case.
But even in a network limited domain, I can see reasons to then prefer `xz` instead due to its better compression ratio and relatively similar decompression speed to gzip.

IMHO, I think it should be ok to not store any suffix at all for the squirrel store.
If I/a user looks _inside_ of the squirrel store URL I think it is not mandatory to _show_ what compression algorithm was used.
The user could/should use the designated driver/metadata that comes bundled with the dataset and let the driver handle the _correct_ decompression.

If you don't agree I still think the `gz` extension doesn't have to be 'hardcoded' into these functions.
This is actually something that confused me when I was looking at the internals of the code base.
So instead, we could use something like:
```python
comp = kwargs.get("compression", "gzip")
comp_to_ext_dict[comp] # just to show the concept
```

With these modifications, it should be possible to utilize different compression modes and make them easily configurable.
I would be very happy to create a PR and contribute to this project!

kai-tub:
# Description

See #59 

Fixes #59 issue

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] Documentation update

# Checklist:

- [X] I have read the [contributing guideline doc](https://squirrel-core.readthedocs.io/en/latest/usage/code_of_conduct.html) (external contributors only)
- [X] I have kept the PR small so that it can be easily reviewed  
- [X] I have made corresponding changes to the documentation
- [X] All dependency changes have been reflected in the pip requirement files. 

# Draft State!

This is a draft PR to make it easier to discuss the different pros and cons of various solutions.
This is **not** in a final state.

I tried to add some test and verify that they pass locally, but the tests spam a lot of
`ValueError: Bucket is requester pays. Set `requester_pays=True` when creating the GCSFileSystem.` and it is hard to tell where these tests/errors are coming from.
The `contributing` guideline provides no further information on how to run the tests.


squirrel/store/squirrel_store.py
```diff
@@ -37,10 +38,16 @@ def get(self, key: str, **kwargs) -> t.Iterator[SampleType]:
...
-        fp = f"{self.url}/{key}.gz"
+        fp = f"{self.url}/{key}"  # Key includes compression suffix
         yield from self.serializer.deserialize_shard_from_file(fp, fs=self.fs, **kwargs)
 
-    def set(self, value: t.Union[SampleType, ShardType], key: t.Optional[str] = None, **kwargs) -> None:
+    def set(
+        self,
+        value: t.Union[SampleType, ShardType],
+        key: t.Optional[str] = None,
+        compression: t.Union[str, None] = "gzip",
```
AlpAribal, line 48: If I use a non-default compression, then I would need to specify it each time I use `set()` or `keys()`. How about making `compression` an init arg?

squirrel/serialization/jsonl.py
```diff
@@ -93,11 +93,11 @@ def serialize_shard_to_file(
...
                 pick a filesystem suitable for `fp`. Defaults to None.
             mode (str): IO mode to use. Passed to :py:meth:`fs.open`. Defaults to "wb".
-            **open_kwargs: Other keyword arguments passed to :py:meth:`fs.open`. `open_kwargs` will always have
-                `compression="gzip"` set.
+            **open_kwargs: Other keyword arguments passed to :py:meth:`fs.open`.
+                `open_kwargs` will always have `compression="infer"` set.
         """
         open_kwargs["mode"] = mode
-        open_kwargs["compression"] = "gzip"
+        open_kwargs["compression"] = "infer"
```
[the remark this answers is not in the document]
kai-tub: Good idea!

kai-tub:
How does the current API look like?
One thing that we might need to pay attention to is the `keys` implementation.
Now we allow the user to explicitly overwrite the `compression` argument during `set/get` calls, but the `keys` method will use the compression set by the init args.
Adding an extra kwarg for `compression` of the `keys` method feels a bit weird.
How should we then _define_ the variable to use `self.keys` if nothing else is provided? `None` is already _reserved_ for disabling the compression.
What is your opinion @AlpAribal?

Also, should we make it more clear that we are 'requiring' the compression method to be registered by `fsspec` for it to function as expected?
Or do we assume that whoever wants to use a _custom_ compression will look up the details anyway?

AlirezaSohofi:
> Now we allow the user to explicitly overwrite the `compression` argument during `set/get` calls, but the `keys` method will use the compression set by the init args.

`keys()` should work will any of the valid compressions (and the absence of file name suffix which means no compression). The reason is that we can't enforce that different processes, concurrent or scattered over time, will use the same compression. The current logic will silently ignore part of the data when reading if the compression is not the same as defined in the `__init__`.

kai-tub:
Ah, I had a different mental model.
In my mind, the SquirrelStore could contain multiple _identical_ variants of the same data with different compression modes.
I was thinking, for example, if we have a dataset with a single image, I could set the compression during init and only get the image with the specific "compression" back (if it exists).
So the `keys` method doesn't _silently_ ignore the other compressed shards but is _set_ via the `init` value.

But yeah, now I see that you are rather thinking that the contents of the SquirrelStore should be invariant to the used compression mode. Meaning that I could compress imageA with `gzip` and imageB with `xz`.
As long as we make it clear that the compression argument in the `init` refers to the "default" compression mode of `set` and has no impact on the `keys` method, I have nothing against it.
My use case could be easily implemented by overwriting the `keys` method with the current version or applying some post-filtering of the keys.

From my view, as a dataset creator, I was thinking that I could provide a single `Dataset` store and let the user choose the "best" compressed mode during the `init` call. Because `None` works great when the Dataset is on a local SSD but `xz/gzip` are better when the data is accessed from a remote machine over a slow network connection.
But as I said, that is something that can be easily implemented in a custom class/wrapper.

A short summary of how the different variables interact with each other before I implement the code:

- `key`: To differentiate between different compression modes, we rely on the suffix of the shard. As a result, we make the entire shard name the `key`, i.e., the key now includes the extension
- `compression` in `init` will now only be used to define the compression mode used during `set` of the `SquirrelStore`
  - Motivation: Only requires setting the preferred compression once and not in all `set` calls
  - Implication: The `key` value provided to `set` will be _extended_ by the compression suffix. The suffix is derived from `self.compression`. This `new_key` will be used during the `serialize_shard_to_file` call.
  - Extensions: We could check if `compression` was present in the kwargs and use that instead of `self.compression` if it was provided. If it isn't provided than `self.compression` is used. This would give the user more control.
  - Alternative: Require that compression is defined in all calls to `set` than we could drop `self.compression` and potentially avoid any confusion about when `self.compression` is used and when it isn't
- `keys` will return _all_ valid files. It will skip over those shards that were compressed in an unsupported format or have an unknown extension.
- `get` the provided `key` includes the suffix of the compression and will be (by default) inferred from within `deserialize_shard_from_file`. We allow a caller to define a different (de)compression mode but it will usually cause conflicts with `key/set` if those methods weren't overwritten.

Comment: With the current design, I would like to rename `compression` in the `init` to `compression_set_default` to make it very clear that this value is only used to add a 'state' to the class.

kai-tub:
PS: I think a nice middle-ground would be to _promote_ the compression keyword argument of `SquirrelStore.set` to its own keyword argument and not be "hidden" behind `**kwargs`.
See my next commit.

AlirezaSohofi:
> I think a nice middle-ground would be to _promote_ the compression keyword argument

Makes sense to me :+1: 

> As long as we add a mini-tutorial/draft-tutorial and show dataset creators how to utilize the different compression methods 

That would be awesome! It would be a very welcome addition to the array of [tutorials](https://github.com/merantix-momentum/squirrel-datasets-core/tree/main/examples). If you have more ideas what to include, feel free to open an issue.

test/test_driver/test_sq_store.py
```diff
@@ -20,14 +20,17 @@ def test_get_iter(dummy_sq_store: SquirrelStore) -> None:
     """Test StoreDriver.get_iter."""
     it = StoreDriver(url=dummy_sq_store.url, serializer=dummy_sq_store.serializer).get_iter().collect()
     assert isinstance(it, list)
+    # TODO: This needs to be fixed to either inject the compression
+    # suffix here, or to disable compression in dummy_sq_store
+    # or to inject the correct compression suffix in `get_sample`
```
kai-tub, line 25: What is your preference for the tests?

AlirezaSohofi:
Hey @kai-tub. This PR has been open for some time, although I think all issues are addressed and even the implementation is almost complete. Is there anything We can help with to finish this PR?

PR #66, opened 2022-07-05
Composable.compose gets source as kwarg, which is equal to self

AlirezaSohofi:
# Description

Not having access to `source` in the `__init__` method when composing iterables makes it difficult to track steps of the pipeline (see #64). 

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:


pzdkn (dismissed):
thanks :) LGTM!

squirrel/iterstream/base.py
```diff
@@ -140,106 +226,14 @@ def monitor(
...
 class _Iterable(Composable):
     """
     A class representing an iterable, which applies the callable `f` to the `source` items and returns the result in
     the :py:meth:`__iter__` method. This is used as the object being passed between steps of the stream.
     """
 
-    def __init__(
-        self, source: t.Iterable, f: t.Callable[..., t.Iterator], *args, _kwa: t.Optional[t.Dict] = None, **kw
-    ):
+    def __init__(self, source: t.Iterable, f: t.Callable[..., t.Iterator], *args, **kw):
```
AlpAribal, line 236: Could you update the docstring as well?

squirrel/iterstream/source.py
```diff
@@ -40,7 +40,7 @@ def __iter__(self) -> t.Iterator:
             yield from self.source
 
 
-class FilePathGenerator(IterableSource):
+class FilePathGenerator(Composable):
     """
     A specialized version of IterableSource that accepts a url without instantiating it eagerly. It simply generates
```
AlpAribal, line 45:
```suggestion
    A specialized version of Composable that accepts a url without instantiating it eagerly. It simply generates
```

AlpAribal (commented):
Neat, nice catch!

It was rather hard to understand the diff since methods were reordered (IMO updating the compose/source_ methods here and doing the ordering in a separate PR, which we could merge to this one, would be easier for review).

pzdkn (approved):
LG!

squirrel/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "0.14.3"
+__version__ = "0.14.4"
```
AlpAribal, line 1: Shall we have a minor bump (since `.source_()` got removed)?

PR #67, opened 2022-07-12
Include source behaviour from polls

pzdkn:
# Description

Include the desired behaviour for passing a source argument when using `compose()`:

- We can define constructor for custom Composable, but must leave out source argument
- We can leave out constructor, and use the base one. 

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:


squirrel/iterstream/base.py
```diff
@@ -40,14 +41,20 @@ def __iter__(self) -> t.Iterator:
     def compose(self, constructor: t.Type[Composable], *args, **kw) -> Composable:
         """Compose the items in the stream with items generated by `constructor`"""
         assert "source" not in kw
-        kw["source"] = self
-        return constructor(*args, **kw)
+        if "source" in inspect.signature(constructor).parameters:
+            assert inspect.signature(constructor).parameters["source"].default is None
```
AlirezaSohofi, line 45: we should disallow having `source` in the parameters, so raise at `45`.

test/test_iterstream/test_stream.py
```diff
@@ -59,16 +59,39 @@ def test_map(samples: t.List[t.Dict]) -> None:
...
 
     class Add1(Composable):
+        # use default constructor, supported by compose
         def __iter__(self):
             for i in self.source:
                 yield i + 1
 
+    class Add2(Composable):
+        # use constructor without source, supported by compose
+        def __init__(self):
```
AlirezaSohofi, line 72: please add a case that has arguments in the init

test/test_iterstream/test_stream.py
```diff
@@ -59,16 +59,39 @@ def test_map(samples: t.List[t.Dict]) -> None:
...
+            super().__init__()
+
+        def __iter__(self):
+            for i in self.source:
+                yield i + 2
+
+    class Add3(Composable):
+        # use constructor with source, not supported by compose
+        def __init__(self, source: t.Union[t.Iterable, t.Callable]):
```
AlirezaSohofi, line 81: please add a case `source=None` in the init. It should fail.

squirrel/iterstream/base.py
AlirezaSohofi, line 45: Please raise an informative exception

AlirezaSohofi (changes requested):

squirrel/iterstream/base.py
pzdkn: If we don't implement a constructor, then the base constructor is used and this one will have a default source set to None. 

As far as I understand we wanted to still support the case of omitting the constructor, wdyt @AlirezaSohofi ?

test/test_iterstream/test_stream.py
pzdkn: passing `None` should be fine right?

squirrel/iterstream/base.py
pzdkn: Not sure if ValueError is the right Exception here

test/test_iterstream/test_stream.py
```diff
@@ -59,16 +59,62 @@ def test_map(samples: t.List[t.Dict]) -> None:
...
+            super().__init__(source)
+
+        def __iter__(self):
+            for i in self.source:
+                yield i + 3
+
+    class Add4(Composable):
```
AlirezaSohofi, line 88: this should fail

squirrel/iterstream/base.py
```diff
@@ -41,7 +41,7 @@ def __iter__(self) -> t.Iterator:
     def compose(self, constructor: t.Type[Composable], *args, **kw) -> Composable:
         """Compose the items in the stream with items generated by `constructor`"""
         assert "source" not in kw
-        if "source" in (param := inspect.signature(constructor).parameters) and param["source"].default is not None:
+        if "__init__" in constructor.__dict__ and "source" in inspect.signature(constructor).parameters:
             raise ValueError(f"source argument in {constructor} should be omitted")
```
AlirezaSohofi, line 45:
```suggestion
            raise ValueError(f"If the class provides as constructor argument has an __init__ method, it should not have the source argument.")
```

AlirezaSohofi (changes requested):
LG, please also update the documentation.

squirrel/iterstream/base.py
```diff
@@ -38,16 +39,28 @@ def __iter__(self) -> t.Iterator:
         pass
 
     def compose(self, constructor: t.Type[Composable], *args, **kw) -> Composable:
-        """Compose the items in the stream with items generated by `constructor`"""
+        """
+        Compose the items in the stream with items generated by `constructor.
```
AlirezaSohofi, line 43:
```suggestion
        Apply the transformation expressed in the `__iter__` method of the `constructor` to items in the stream.
```

PR #72, opened 2022-07-29
Unifies folder-creation behaviour when instantiation SquirrelStore

pzdkn:
# Description
Currently local and GCS filesystems show different behaviour, when SquirrelStore is initialized. In local filesystems, a root directory is created, whereas for GCS a creation is not possible.We solve this by creating the directory only when the first shard is written. 

Another related issue fixed in this PR, is that previously if a directory does not exist and the `clean` flag is set to true, an error is thrown.  The directory removal is now skipped, when the dir does not exist.

## Type of change

- [ ] Documentation update

# Checklist:


test/test_driver/test_sq_store.py
```diff
@@ -1,13 +1,25 @@
...
 
 from squirrel.driver import StoreDriver
 from squirrel.iterstream import FilePathGenerator
+from squirrel.serialization import MessagepackSerializer
+from squirrel.store import SquirrelStore
 
-if TYPE_CHECKING:
-    from squirrel.store import SquirrelStore
+
+def test_non_existent_path() -> None:
```
AlpAribal, line 12: I recommend using the `test_path` fixture here (defined in `squirrel.integration_test.shared_fixtures`). It makes sure we try both local and GCS paths. You can check out the `test_catalog_saveandload` test for an example.

Note that `test_path` is a `str` and might contain filesystem protocol, so cannot use pathlib operations on it, would need to append the path like `f"{test_path}/non_existing"`.

It would be nice to also test the behavior when we want to create a store in an *existing* path, for completeness.

test/test_driver/test_sq_store.py
```diff
@@ -1,13 +1,25 @@
...
+    """Test that a non-existent path returns an empty list."""
+    temp_path_1 = pl.Path.cwd() / "tmp" / "non_existent_path_1"
```
AlirezaSohofi, line 14: please test `gs` paths too.

AlirezaSohofi (changes requested):

test/test_driver/test_sq_store.py
pzdkn: @AlirezaSohofi , what is the expected behaviour for gcs path though? Afaik, we can't create a folder structure there.

AlpAribal: After writing at least one shard, the behavior should be the same for both filesystems. When writing a file (say, at `/path/dir/file`), fsspec adds some metadata to "folder" paths (/path and /path/dir) and later on it can infer that there is a directory at these paths.

So, we can write a shard in the test and then do the very same test for local and gcs filesystems (via the `test_path` fixture I mentioned above)

pzdkn: Hm but we want the directories to be created when the Store is initialized not manually beforehand by writing to the store, right @AlpAribal ?

squirrel/store/squirrel_store.py
```diff
@@ -57,6 +57,10 @@ def set(self, value: t.Union[SampleType, ShardType], key: t.Optional[str] = None
         if key is None:
             key = get_random_key()
         fp = f"{self.url}/{key}.gz"
+
+        if not self.fs.exists(self.url):
+            self.fs.makedirs(self.url, exist_ok=True)
```
AlirezaSohofi, line 62:
```suggestion
        self._create_dir_if_does_not_exists():
```

AlirezaSohofi: Please introduce `self._dir_created = False` and avoid calling `self.fs.exists` for every set operation

AlirezaSohofi (changes requested):
Looks good, just a minor comment to address.

test/test_driver/test_sq_store.py
```diff
@@ -1,13 +1,31 @@
...
+
+    cleaned_store = SquirrelStore(url=dummy_sq_store.url, serializer=MessagepackSerializer(), clean=True)
+    assert not cleaned_store.fs.exists(dummy_sq_store.url)
+    assert len(list(cleaned_store.keys())) == 0
+
+
+@pytest.mark.parametrize("clean", [True, False])
+def test_store_creation_empty_path(test_path: str, clean: bool) -> None:
+    """Test store creation with an empty path."""
+    store = SquirrelStore(url=test_path, serializer=MessagepackSerializer(), clean=clean)
```
AlpAribal, line 26: In case we change how we implement `test_path`, e.g. we return an existing path.

After importing
```
```

```suggestion
    fs = get_fs_from_url(path)
    assert not fs.exists(test_path)
    store = SquirrelStore(url=test_path, serializer=MessagepackSerializer(), clean=clean)
```

test/test_driver/test_sq_store.py
```diff
@@ -40,3 +58,6 @@ def test_get_iter(dummy_sq_store: SquirrelStore) -> None:
 def test_filepathgenerator(dummy_sq_store: SquirrelStore, num_samples: int) -> None:
     """Test FilePathGenerator."""
     assert len(FilePathGenerator(url=dummy_sq_store.url).collect()) == num_samples
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        assert len(FilePathGenerator(url=tmpdir).collect()) == 0
```
AlpAribal, line 63: :rocket:

AlpAribal (changes requested):
Thanks for aligning different filesystems! Ali's comment makes a lot of sense, `fs.exists()` calls might add up if the store is huge

squirrel/store/filesystem.py
```diff
@@ -41,9 +41,9 @@ def __init__(
         self.storage_options = storage_options
         self.serializer = serializer
         self.fs = get_fs_from_url(self.url, **self.storage_options)
-        if clean:
+        self._dir_exists = self.fs.exists(self.url)
+        if clean and self._dir_exists:
             self.fs.rm(self.url, recursive=True)
```
AlpAribal, line 46: At this point, the dir does not exist anymore, so we should set `self._dir_exists=False`. This bug can be validated by adding a `store.set(...)` at the end of the newly-added store creation tests.

test/test_driver/test_sq_store.py
```diff
@@ -1,13 +1,34 @@
...
+
+def test_store_creation(dummy_sq_store: SquirrelStore, num_samples: int) -> None:
+    """Test store creation with non empty path."""
+    uncleaned_store = SquirrelStore(url=dummy_sq_store.url, serializer=MessagepackSerializer(), clean=False)
+    assert len(list(uncleaned_store.keys())) == num_samples
+
+    cleaned_store = SquirrelStore(url=dummy_sq_store.url, serializer=MessagepackSerializer(), clean=True)
+    assert not cleaned_store.fs.exists(dummy_sq_store.url)
+    assert len(list(cleaned_store.keys())) == 0
+
```
AlpAribal, line 22: Let's also try writing to the store here

test/test_driver/test_sq_store.py
```diff
@@ -1,13 +1,34 @@
...
+
+@pytest.mark.parametrize("clean", [True, False])
+def test_store_creation_empty_path(test_path: str, clean: bool) -> None:
+    """Test store creation with an empty path."""
+    fs = get_fs_from_url(test_path)
+    assert not fs.exists(test_path)
+    store = SquirrelStore(url=test_path, serializer=MessagepackSerializer(), clean=clean)
+    assert not fs.exists(test_path)  # Directory should not be created at initialization
+    assert list(store.keys()) == []
```
AlpAribal, line 31: Let's also try writing to the store here

AlpAribal (changes requested):

squirrel/store/filesystem.py
pzdkn: How about when someone passes a non-empty directory at initialisation ?

pzdkn: that's a good catch, thanks!

AlpAribal (dismissed):
LGTM! Please make at least a minor version bump since we remove a public method.

AlirezaSohofi (dismissed):
LG :+1:

AlpAribal (approved):
LGTM, thank you for the meticulous work!

pzdkn:
Thanks for the great tips!

PR #76, opened 2022-08-03
Nvidia DALI external source integration

axkoenig:
# Description

**Motivation**
Squirrel is a fast data loading framework, and Nvidia DALI is a fast, gpu-accelerated library for complex ML workloads such as run-time augmentations. The aim is to provide users with an intuitive interface to use Squirrel as a backend for Nvidia DALI. 

**Context**
For more context, check out these internal benchmarks below. Running the Squirrel pipeline without any augmentations is approx. 33k samples / sec. If you use Squirrel as an external source and an affine image augmentation from DALI you can reach approx. 28k samples / sec. This suggests that DALI can make use full use of Squirrel's speed as the data loading speed is almost not slowed down by the run-time augmentations (33k vs 28k). DALI brings two things to the table: you can augment data in batches and not one-by-one as is necessary with the other frameworks, and you can do it on the GPU. 

Screen Shot 2022-08-03 at 15 36 30

**Code Design**
- DALI comes with a concept called "pipeline" ([docs](https://docs.nvidia.com/deeplearning/dali/main-user-guide/docs/pipeline.html)), that defines how data should be read and transformed by DALI.  
- We use the `external_source` data reader API in the DALI pipeline, which we can provide with a modified Squirrel Iterable, the `squirrel.iterstream.DaliExternalSource`. 
- As suggested by Nvidia DALI staff, I benchmarked loading the samples one-by-one and let DALI do the batching. It turned out that batching in Squirrel was much faster (18.2k sps vs 32.2 sps). This suggests that DALI profits from the async loading in Squirrel here. 
- As suggested by Nvidia DALI staff, I tried using [parallel external source](https://docs.nvidia.com/deeplearning/dali/user-guide/docs/examples/general/data_loading/parallel_external_source.html?highlight=parallel%20source), which is multi-proc dataloading by DALI. As stated in their docs, DALI prefers single samples (un-batched) here to let DALI handle the multi-proc logic of parallel data fetching. The problem here is that DALI would like a `Callable` external source here, `Iterables` are not allowed for parallel fetching. While this is technically possible (e.g. fit your dataset in one shard and then access the items by their keys, i.e. shard names), indexability is not straightforward and not yet integrated in squirrel. Since DALI already nearly makes use of Squirrel's full performance, we don't see that DALI could speed things up here. But it's worth investigating once the feature is implemented in Squirrel. 
- There was no performance increase by returning `cupy` arrays on the GPU to the `external_source` reader. Numpy was slightly faster, so users are advised to return numpy arrays in their collation function.

**Usage Pattern**
- users will simply turn their iterable into an external source with the iterstream API. 
```python
# define a dummy pipeline
@pipeline_def 
def pipeline(it: DaliExternalSource, device: str) -> Tuple[DataNode]:
    img, label = fn.external_source(source=it, num_outputs=2, device=device)
    enhanced = fn.brightness_contrast(img, contrast=2) # do other augmentations here
    return enhanced, label
    
it = squirrel_iterator.to_dali_external_source(batch_size, my_collation_fn)
pipe = pipeline(it, device, batch_size=BATCH_SIZE)
pipe.build()

loader = DALIGenericIterator([pipe], ["img", "label"])
for item in loader: 
    # ... 
```

**Things to Discuss**
1. I tried turning the iterstream into a `DALIGenericIterator` directly and abstract the above code away, but in my mind that does not make a lot of sense, as DALI users are used to the above API and we are really just an external source. The user will need to define their custom pipeline anyway for their use-case, so I don't see a big benefit of abstracting the below code away into a squirrel functionality - possibly adding some assumptions here and there and thereby limiting the original functionality of DALI (wdyt @AlirezaSohofi ?). 
2. We would need to find out if the `self.i` and `self.n` parameters need to be set for the external source as indicated [here](https://docs.nvidia.com/deeplearning/dali/user-guide/docs/examples/general/data_loading/external_input.html). For now, it seems to work out of the box, but maybe for more complex use-cases these variables are needed for DALI to keep track of the loaded samples. Sidenote: Currently `DaliExternalSource` could also simply be replaced with `squirrel_iterable.batched(bs, fn)`, but I assume that `self.i` and `self.n` are needed somehow (input from NVIDIA needed here), so it's useful to have `DaliExternalSource` where we can add more features.
3. Please check out the `test_to_dali_external_source_gpu_multi_epoch`. After iterating over Squirrel's generator once the iterable is empty. Hence after each epoch we need to create a new `DALIGenericIterator`. Afaik this is also how e.g. Pytorch Lightning handles it. Let me know if that sounds ok, or if we need to loop over the data. 
4. Tests & Requirements: Note that I added pytests for the code, but did not update the requirements accordingly, because the CI currently doesn't run GPU tasks. Moreover, we won't ask users to install DALI for now (also, there are many different versions for different cuda drivers), so we assume people will prefer installing themselves. The `DaliExternalSource` doesn't depend on any DALI code, so the DALI install is technically not required.

## Type of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -2,5 +2,7 @@
 testpaths =
     test
     squirrel/integration_test
-addopts = -p no:warnings -v
+# ignore dali tests for now because we need CUDA + DALI installed and a GPU for 
+# these tests to run, but currently our testing image is CPU-only
+addopts = -p no:warnings -v --ignore=test/test_iterstream/test_dali_composables.py
 norecursedirs = '.*', 'build', 'dist', 'CVS', '_darcs', '{arch}', '*.egg'
--- a/squirrel/iterstream/base.py
+++ b/squirrel/iterstream/base.py
@@ -257,8 +257,10 @@ class Composable:
         return self.compose(iterstream.torch_composables.TorchIterable)
 
     def to_dali_external_source(self, batch_size: int, collation_fn: t.Callable[[t.List], t.List]) -> Composable:
-        """Convert the stream to a Nvidia DALI external source that is directly usable by nvidia.dali.fn.external_source
-        in a DALI pipeline."""
+        """
+        Convert the stream to a Nvidia DALI external source that is directly usable by nvidia.dali.fn.external_source
+        in a DALI pipeline.
+        """
         return self.compose(iterstream.dali_composables.DaliExternalSource, batch_size, collation_fn)
 
 
--- a/squirrel/iterstream/base.py
+++ b/squirrel/iterstream/base.py
@@ -256,6 +256,11 @@ class Composable:
         """Convert the stream to a torch iterable."""
         return self.compose(iterstream.torch_composables.TorchIterable)
 
+    def to_dali_external_source(self, batch_size: int, collation_fn: t.Callable[[t.List], t.List]) -> Composable:
+        """Convert the stream to a Nvidia DALI external source that is directly usable by nvidia.dali.fn.external_source
+        in a DALI pipeline."""
+        return self.compose(iterstream.dali_composables.DaliExternalSource, batch_size, collation_fn)
+
 
 class _Iterable(Composable):
     """
--- a/squirrel/iterstream/dali_composables.py
+++ b/squirrel/iterstream/dali_composables.py
@@ -1,0 +1,67 @@
+from __future__ import annotations
+
+from collections import defaultdict
+from typing import Any, Callable, Dict, List
+
+import numpy as np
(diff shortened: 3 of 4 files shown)


axkoenig:
I have read the CLA Document and I hereby sign the CLA

axkoenig:
Closing since Squirrel works as a backend for DALI out of the box - which wasn't clear at the time of writing this PR. Hence, only some docs on how DALI + Squirrel work together are required. New PR with documentation will be out soon.

PR #77, opened 2022-08-04
Warn when creating driver that points to an empty directory

pzdkn:
# Description
A warning is shown when we iterate over a driver, that points to an empty or non-existent directory. 
## Type of change

- [ ] Documentation update

# Checklist:


squirrel/fsspec/fs.py
```diff
@@ -45,3 +45,12 @@ def create_dir_if_does_not_exist(fs: fsspec.spec.AbstractFileSystem, path: str)
...
         fs.mkdir(path)
+
+
+def validate_path(url: str) -> None:
+    """Validate that path exists and that path points to a non-empty directory"""
+    fs = get_fs_from_url(url)
+    if not fs.exists(url):
+        raise FileNotFoundError(f"Path {url} does not exist")
+    if fs.isdir(url) and len(fs.find(url)) == 0:
+        raise ValueError(f"Path {url} is an empty directory")
```
AlirezaSohofi, line 56: we should only [warn](https://docs.python.org/3/library/warnings.html#warnings.warn) the user, not to raise an exception.

squirrel/driver/store_driver.py
```diff
@@ -28,6 +29,8 @@ def __init__(self, url: str, serializer: SquirrelSerializer, **kwargs) -> None:
             **kwargs: Keyword arguments to pass to the super class initializer.
         """
         super().__init__(**kwargs)
+
+        validate_path(url)
```
AlirezaSohofi, line 33: The warning should be generated when the user tries to read from a store that is empty. It's a normal use case where the user instantiates a driver that points out to an empty or non-existing uri, with the intend to write to it.

AlirezaSohofi (changes requested):

squirrel/driver/store_driver.py
AlpAribal: Drivers are not able to write by definition, shall we align on this?

AlirezaSohofi: This has been so far recommended, but not enforced. Having the property `store` in the driver implied that you can use driver api to instantiate something that can write. If we enforce this, we should remove the `store` property too imo.

squirrel/fsspec/fs.py
```diff
@@ -45,3 +47,10 @@ def create_dir_if_does_not_exist(fs: fsspec.spec.AbstractFileSystem, path: str)
     """If the given path does not exist, creates a directory at it."""
     if not fs.exists(path):
         fs.mkdir(path)
+
+
+def validate_path(url: str) -> None:
+    """Validate that path exists and that path points to a non-empty directory"""
+    fs = get_fs_from_url(url)
+    if fs.isdir(url) and len(fs.find(url)) == 0:
```
AlirezaSohofi, line 55: `fs.find` will try to list all the files, this adds an unacceptable overhead to the majority of cases. 

`fs.isdir` does not work for gcs if the data is not written by gcsfs (@AlpAribal  correct me if I am wrong).

I think it not possible to avoid overhead if this is supposed to handle general cases. Since it is only used in `SquirrelStore`, which is always a flat directory of files, we don't need to consider nested directory structures.

AlirezaSohofi: I suggest to move the validation to `FilePathGenerator` instead of the driver. Some advantages:

- it is used by many different drivers and even without drivers, so it's more widely useful
- we won't add any overhead

If FPG does not yields any item, then we can warn the user.

test/test_driver/test_msgpack.py
```diff
@@ -10,6 +11,22 @@
...
 
+def test_invalid_url() -> None:
+    """Test if an error is raised when an invalid url is passed"""
+    # empty directory
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        with pytest.warns(UserWarning):
+            _ = MessagepackDriver(url=tmp_dir).get_iter()
+
+    # directory containing only empty directories
+    with tempfile.TemporaryDirectory() as tmp_dir:
```
AlirezaSohofi, line 22: what should happen if the directory does not exists?

squirrel/fsspec/fs.py
```diff
@@ -45,3 +47,10 @@ def create_dir_if_does_not_exist(fs: fsspec.spec.AbstractFileSystem, path: str)
     """If the given path does not exist, creates a directory at it."""
     if not fs.exists(path):
         fs.mkdir(path)
+
+
+def validate_path(url: str) -> None:
```
AlirezaSohofi, line 52: more descriptive name please

AlirezaSohofi (changes requested):

AlpAribal:
Shall we base this branch onto #72, there are some common changes in both, it might be safer to see how the changes combined work together.

squirrel/fsspec/fs.py
AlpAribal: > `fs.find` will try to list all the files, this adds an unacceptable overhead to the majority of cases.

Correct :+1: 

> `fs.isdir` does not work for gcs if the data is not written by gcsfs (@AlpAribal correct me if I am wrong).

That is also correct :+1: 
 
> I suggest to move the validation to FilePathGenerator instead of the driver.

We can maybe use [more_itertools.peekable()](https://more-itertools.readthedocs.io/en/latest/api.html#more_itertools.peekable) to check if the iterable is empty or not.

test/test_driver/test_msgpack.py
AlpAribal: This is handles in #72, I really think we should merge these two PRs

AlirezaSohofi:
> Shall we base this branch onto #72, there are some common changes in both, it might be safer to see how the changes combined work together.

agree, can you please rebase this branch please @pzdkn?

pzdkn:
I have rebased and tests for both branches pass

squirrel/iterstream/source.py
```diff
@@ -95,8 +97,17 @@ def __iter__(self) -> t.Iterator[str]:
                         urls.extend(d)
         else:
             for url in urls:
+                if not self._returned_url:
+                    self._returned_url = True
```
AlpAribal, line 101: We need the same check for the `nested` case as well. I think, right now, we will get an empty warning for a dir containing only dirs but no files

test/test_driver/test_msgpack.py
```diff
@@ -10,6 +11,34 @@
...
+        with pytest.warns(UserWarning):
+            _ = MessagepackDriver(url=tmp_dir).get_iter().collect()
+
+    # directory containing only empty directories
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        with tempfile.TemporaryDirectory(dir=tmp_dir) as sub_dir:
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
+                _ = MessagepackDriver(url=tmp_dir).get_iter().collect()
```
AlpAribal, line 27: Why do we catch/filter warnings here?

test/test_driver/test_msgpack.py
```diff
@@ -10,6 +11,34 @@
...
+                _ = MessagepackDriver(url=sub_dir).get_iter().collect()
+
+    # test invalid url
+    invalid_url = "invalid_url"
+    with pytest.warns(UserWarning):
+        _ = MessagepackDriver(url=invalid_url).get_iter().collect()
+
+    # test valid url containing samples
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
```
AlpAribal, line 38: Why do we catch/filter warnings here?

test/test_driver/test_msgpack.py
```diff
@@ -10,6 +11,34 @@
 from squirrel.store import SquirrelStore
 
 
+def test_invalid_url(local_msgpack_url: URL) -> None:
```
AlpAribal, line 14: Would rename this test, it does not only use invalid urls.

AlpAribal: We need to add `nested` cases here, perhaps via `parametrize`

AlpAribal (changes requested):

squirrel/iterstream/source.py
pzdkn: Do we want to cover the nested case though? We are not reading in a nested way in the first place.

test/test_driver/test_msgpack.py
pzdkn: This is to check that when there are items in the store, then no warnings shall be printed. 
This `.simplefilter` asserts that no warning is shown:

https://stackoverflow.com/questions/45671803/how-to-use-pytest-to-assert-no-warning-is-raised

pzdkn: But it is in fact unintuitive, since a directory containing empty directories will no yield warnings

squirrel/iterstream/source.py
AlirezaSohofi: FilePathGenerator is not only menat to be used in `SquirrelStore`, so it should be general and and able to handle nested cases too.

squirrel/iterstream/source.py
```diff
@@ -89,14 +91,25 @@ def __iter__(self) -> t.Iterator[str]:
                             future = AsyncContent(url, self.fs.ls, pool)
                             dirs.append(future)
                         else:
-                            yield f"{self.protocol}{url}"
+                            if not self._returned_url:
+                                self._returned_url = True
+                                yield f"{self.protocol}{url}"
```
AlirezaSohofi, line 96:
```suggestion
                            yield f"{self.protocol}{url}"
```

squirrel/iterstream/source.py
```diff
@@ -89,14 +91,28 @@ def __iter__(self) -> t.Iterator[str]:
...
+                            if not self._returned_file_url:
+                                self._returned_file_url = True
                             yield f"{self.protocol}{url}"
                     if (len(dirs) >= self.max_dirs and len(urls) < self.max_keys) or len(urls) == 0 and dirs:
                         d = dirs.pop(0).value()
                         urls.extend(d)
         else:
             for url in urls:
+                if not self._returned_file_url and not self.fs.isdir(path=url):
+                    self._returned_file_url = True
```
AlirezaSohofi, line 103:
```suggestion
                if not self._returned_file_url:
                    self._returned_file_url = True

```

AlirezaSohofi: why do we need ` not self.fs.isdir(path=url)`?

AlirezaSohofi (changes requested):

PR #78, opened 2022-08-11
Bugfix - Register Torch Composables

axkoenig:
# Description

Bug: The problem regards `split_by_worker_pytorch` `split_by_rank_pytorch` and `to_torch_iterable`. Checkout the example
```python
# comment in below line and it works
# from squirrel.iterstream.torch_composables import TorchIterable

SOURCE = ["key1","key2","key3","key4"]
IterableSource(SOURCE).to_torch_iterable().map(print).join()
```

Fails with this error.

```bash
Traceback (most recent call last):
  File "/Users/koenig/Documents/projects/sandbox/test.py", line 13, in <module>
    doesnt_work()
  File "/Users/koenig/Documents/projects/sandbox/test.py", line 6, in doesnt_work
    IterableSource(SOURCE).to_torch_iterable().map(print).join()
  File "/Users/koenig/Documents/projects/sandbox/.sandbox/lib/python3.9/site-packages/squirrel/iterstream/base.py", line 257, in to_torch_iterable
    return self.compose(iterstream.torch_composables.TorchIterable)
AttributeError: module 'squirrel.iterstream' has no attribute 'torch_composables'
```
I think the problem is that `iterstream.__init__.py` isn’t registering `torch_composables.py`. Adding the below fixes the problem.

This was previously not caught by our tests because the import statement `from squirrel.iterstream.torch_composables import SplitByRank, SplitByWorker, TorchIterable` in `test_iterstream.test_torch_composable` registers the modules. 

## Type of change

- [ ] Documentation update

# Checklist:


squirrel/iterstream/__init__.py
```diff
@@ -6,5 +6,14 @@
 
 from squirrel.iterstream.base import Composable
 from squirrel.iterstream.source import FilePathGenerator, IterableSamplerSource, IterableSource
+from squirrel.iterstream.torch_composables import SplitByWorker, SplitByRank, TorchIterable
 
-__all__ = ["Composable", "IterableSource", "FilePathGenerator", "IterableSamplerSource"]
+__all__ = [
```
pzdkn, line 11: Thanks for catching this! I think the `SplitByWorker` etc doesn't have to be in` __all__`, right?
Only importing `torch_composables` should be enough for visibility, but I am not sure if this is best practice.

axkoenig: good point, maybe @AlpAribal or @AlirezaSohofi can help out here?

AlpAribal: I guess it wasn't the best design to separate `Composable` and torch composable into two separate modules since both depend on each other.

I do not think it is necessary to make torch composables accessible at the iterstream level, so I agree with not having them in `__all__`. I also don't like importing the `torch_composables` without really using it. How about lazily importing corresponding torch composables inside the `Composable.split_by_xxx()` and `Composable.to_torch_iterable()` methods?

axkoenig: Wouldn't importing inside the to_torch_iterable() potentially be slow because this composable is executed for every item, right?

AlpAribal: Python caches imported modules, I do not think this will be a problem

axkoenig: understood, I implemented your suggestion and it works, let's merge this?

AlpAribal (approved):
LGTM, thanks for discovering and fixing!

PR #79, opened 2022-08-16
Upgrade infra to py3.9

AlpAribal:
# Description

Bumps the python version in the base image to 3.9 and refreezes requirements with py3.9.

All frozen requirements were also updated to their latest suitable version (not a must, can take back).

I started with an empty requirements file so that the hashes were not reused. In this case, pip-compile generates hashes for all binaries/wheels of the package, which is the reason for the huge diff in reqs.txt

No changes were made in the python version in `setup.py` since we still want to support py3.8

## Type of change

- [ ] Documentation update

# Checklist:


requirements.txt
```diff
@@ -2,15 +2,16 @@
 # This file is autogenerated by pip-compile with python 3.9
 # To update, run:
 #
-#    pip-compile --allow-unsafe --generate-hashes --no-annotate --no-emit-index-url --output-file=requirements.txt --strip-extras requirements.azure.in requirements.dask.in requirements.dev.in requirements.doc.in requirements.gcp.in requirements.in requirements.s3.in requirements.torch.in requirements.zarr.in
+#    pip-compile --allow-unsafe --generate-hashes --no-annotate --no-emit-index-url --output-file=requirements.txt --resolver=backtracking --strip-extras requirements.azure.in requirements.dask.in requirements.dev.in requirements.doc.in requirements.gcp.in requirements.in requirements.s3.in requirements.torch.in requirements.zarr.in
```
AlpAribal, line 5: Backtracking resolver (resolver used by pip since v20.3) is more robust than the default one

AlpAribal (commented):
We need to trigger the main build (using this branch) to build a base image with python3.9 inside. Then, the branch build should succeed.

winfried-loetzsch (approved):
LG! I triggered the main build

PR #80, opened 2022-08-18
Add a mechanism to prevent unintentionally writing duplicate data to the store

pzdkn:
# Description
Currently, re-running some code that was used to create a store (e.g. a data transformer flyte task in chameleon) will  silently extend that store with duplicated data. This can easily lead to fuck ups and kill reproducibility, as a e.g. a config that refers to a catalog name and version suddenly contains different data, because someone was re-running the same data creation code (e.g. if flyte caching somehow breaks, and a task that creates the store gets re-triggered by flyte it makes the store completely useless)

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:


test/test_driver/test_sq_store.py
```diff
@@ -5,13 +5,59 @@
...
-from squirrel.iterstream import FilePathGenerator
+from squirrel.iterstream import FilePathGenerator, IterableSource
 from squirrel.serialization import MessagepackSerializer
 from squirrel.store import SquirrelStore
 from squirrel.fsspec.fs import get_fs_from_url
 from squirrel.integration_test.helpers import SHAPE, get_sample
 
 
+@pytest.mark.parametrize("exist_ok_while_clean", [True, False], "clean_while_exist_ok", [True, False])
+def test_squirrel_overwrite_when_creating(
```
pzdkn, line 16: Should we throw an error when the url is non-empty and add the `exist_ok` flag during creation?

test/test_driver/test_sq_store.py
```diff
@@ -5,13 +5,59 @@
...
+    # No error should be thrown when instantiating a store with a non-empty url and clean is set to True
+    try:
+        SquirrelStore(
+            url=dummy_sq_store.url, serializer=MessagepackSerializer(), clean=True, exist_ok=exist_ok_while_clean
+        )
+    except ValueError:
+        pytest.fail("Expected no error when instantiating from non-empty url and clean is set to True")
+
+
+def test_squirrel_overwrite_when_writing(dummy_sq_store: SquirrelStore, num_samples: int) -> None:
```
pzdkn, line 39: Or should we add an `exist_ok` when writing? Then we would need sample.map(lambda x: store(set(x, exist_ok=True))

test/test_driver/test_sq_store.py
```diff
@@ -5,13 +5,59 @@
...
+    """Test overwrite when setting."""
+    # Error should be thrown when writing to an existing store
+    samples = IterableSource(dummy_sq_store.keys()).async_map(dummy_sq_store.get)
+    with pytest.raises(ValueError):
+        IterableSource(samples).async_map(dummy_sq_store.set)
+
+    # Error should not be thrown when writing an existing store and exit_ok is set to True
+    store = SquirrelStore(url=dummy_sq_store.url, serializer=MessagepackSerializer(), exist_ok=True)
```
pzdkn, line 47: One actually just needs to set exist_ok=True to prevent unintentional overwrites, but this is not user-friendly/intuitive?

test/test_driver/test_sq_store.py
```diff
@@ -5,13 +5,59 @@
...
+    dummy_sq_store: SquirrelStore, exist_ok_while_clean: bool, clean_while_exist_ok: bool
+) -> None:
+    """Test overwriting behaviour when creating a store."""
+    # Error should be thrown when instantiating a store with a non-empty url
+    with pytest.raises(ValueError):
+        SquirrelStore(url=dummy_sq_store.url, serializer=MessagepackSerializer())
+    # No error should be thrown when having the exist_ok flag set to True
+    try:
+        SquirrelStore(
+            url=dummy_sq_store.url, serializer=MessagepackSerializer(), exist_ok=True, clean=clean_while_exist_ok
```
AlirezaSohofi, line 26: `exist_ok` should be the default imo. wdyt @AlpAribal ?

PR #83, opened 2022-08-31
Fix storage options

axkoenig:
# Description

The storage options for the filesystem were not correctly passed down at two places in the code, which resulted in users being unable to set e.g. `storage_options = {"requester_pays": True}`

## Type of change

- [x] Documentation update

# Checklist:


squirrel/driver/store_driver.py
```diff
@@ -19,17 +19,21 @@ class StoreDriver(MapDriver):
...
         Args:
             url (str): the url of the store
             serializer (SquirrelSerializer): serializer to be passed to SquirrelStore
+            storage_options (Optional[Dict[str, Any]]): a dict with keyword arguments passed to filesystem initializer
             **kwargs: Keyword arguments to pass to the super class initializer.
         """
         super().__init__(**kwargs)
         self.url = url
         self.serializer = serializer
+        self.storage_options = storage_options if storage_options is not None else {}
```
AlpAribal, line 36: Just to demonstrate the shorthand for this, do not need to apply
```suggestion
        self.storage_options = storage_options or {}
```

squirrel/store/filesystem.py
```diff
@@ -87,11 +87,15 @@ def keys(self, nested: bool = True, **kwargs) -> t.Iterator[str]:
             nested (bool): Whether to return paths that are not direct children of the root directory. If True, all
                 paths in the store will be yielded. Otherwise, only the top-level paths (i.e. direct children of the
                 root path) will be yielded. This option is passed to FilePathGenerator initializer. Defaults to True.
-            **kwargs: Other keyword arguments passed to the FilePathGenerator initializer.
+            **kwargs: Other keyword arguments passed to the FilePathGenerator initializer. Any kwargs you provide
+                here and are also present in the FilesystemStore's storage_options will be overriden with kwargs.
```
AlpAribal, line 91:
```suggestion
            **kwargs: Other keyword arguments passed to the FilePathGenerator initializer. If a key is present in
                both `kwargs` and `self.storage_options`, the value from `kwargs` will be used.
```

AlpAribal (approved):

PR #85, opened 2022-09-02
Fix file driver storage options

axkoenig:
# Description

The storage options of the file system were not correctly exposed in the FileDriver, which resulted in errors on `requester_pays=True` buckets.  

## Type of change

- [ ] Documentation update

# Checklist:


squirrel/driver/file_driver.py
```diff
@@ -27,13 +29,17 @@ def open(self, mode: str = "r", create_if_not_exists: bool = False, **kwargs) ->
...
             (IO) File handler for the file at `self.path`.
         """
         from squirrel.fsspec.fs import get_fs_from_url
 
-        fs = get_fs_from_url(self.path, storage_options={"auto_mkdir": create_if_not_exists})
+        storage_options = self.storage_options.copy()
+        storage_options.update({"auto_mkdir": create_if_not_exists})
+
+        fs = get_fs_from_url(self.path, **storage_options)
         return fs.open(self.path, mode=mode, **kwargs)
```
AlpAribal, line 45: Directly using `fsspec.open(self.path, mode=mode, **{"auto_mkdir": create_if_not_exists, **kwargs})`, instead of first instantiating our fs and then opening with it, might be a cleaner option here.

axkoenig: afaik `get_fs_from_url` does some path splitting under the hood, wouldn't this be required to get the correct path? or does `fsspec.open()` already take care of correctly parsing `"gs://some_path"` paths. If yes, I wonder why we have the `get_fs_from_url ` in the first place

AlpAribal: Yes, `fsspec.open()` takes care of figuring out the correct filesystem. I would say maybe now we have a better grasp of fsspec :man_shrugging:

axkoenig:
integrated your comments. I tested locally and directly instantiating the fsspec filesystem also works. I wonder why we have `get_fs_from_url` in the first place then, if we can directly instantiate the filesystem without problems. @AlirezaSohofi do you expect that not using `get_fs_from_url` could break anything or is this change legitimate?

squirrel/driver/file_driver.py
```diff
@@ -27,13 +30,12 @@ def open(self, mode: str = "r", create_if_not_exists: bool = False, **kwargs) ->
...
             **kwargs: Keyword arguments that are passed to the `filesystem.open()` method.
 
         Return:
             (IO) File handler for the file at `self.path`.
         """
-        from squirrel.fsspec.fs import get_fs_from_url
-
-        fs = get_fs_from_url(self.path, storage_options={"auto_mkdir": create_if_not_exists})
-        return fs.open(self.path, mode=mode, **kwargs)
+        kwargs = {**self.storage_options, **{"auto_mkdir": create_if_not_exists}, **kwargs}
```
AlpAribal, line 40: This also works
```suggestion
        kwargs = {**self.storage_options, "auto_mkdir": create_if_not_exists, **kwargs}
```

AlpAribal (approved):
Previously, we were manually checking the file path for the protocol and returning our custom GCSFilesystem if the protocol was `gs`. I think this was the main reason for the `get_fs_from_url()` function. Now, we register our filesystem in `setup.py` (see `setup(..., entry_points=...)`), so fsspec automatically uses our fs type.

PR #90, opened 2022-09-19
Peng add kwargs to map

pzdkn:
# Description

Currently to pass additional arguments to a mapped function, one needs wrap the function into `functools.partial`. This is not straightforward, and not very convenient. We instead wrap the kwargs inside the `map`, `async_map` and `dask_map`

Fixes # issue

## Type of change

- [x] Documentation update

# Checklist:


squirrel/iterstream/base.py
```diff
@@ -61,9 +62,10 @@ def source_(self, source: t.Union[t.Iterable, t.Callable]) -> Composable:
         self.source = source
         return self
 
-    def map(self, callback: t.Callable) -> _Iterable:
+    def map(self, callback: t.Callable, **kw) -> _Iterable:
         """Applies the `callback` to each item in the stream"""
-        return self.to(map_, callback)
+        partial_callback = partial(callback, **kw)
```
AlirezaSohofi, line 67:
```suggestion
    def map(self, callback: t.Callable, *arg, **kw) -> _Iterable:
        """Applies the `callback` to each item in the stream"""
        partial_callback = partial(callback, *arg, **kw)
```

AlirezaSohofi: same should be possible for other similar methods

test/test_iterstream/test_stream.py
```diff
@@ -52,8 +52,11 @@ def __iter__(self):
 def test_map(samples: t.List[t.Dict]) -> None:
     """Test map"""
     res_1 = IterableSource(samples).map(lambda sample: _f(sample, 3)).map(lambda sample: sample["label"]).collect()
-
     assert all(i == 3 for i in res_1)
+    res_2 = IterableSource(res_1).map(lambda sample, offset: sample + offset, offset=1).collect()
+    assert all(i == 4 for i in res_2)
+    res_3 = IterableSource(res_2).map(multiply, factor=2).collect()
```
AlirezaSohofi, line 58: please add test the case where we only receive `*args`, or only `**kwargs`, or both.

AlirezaSohofi: please also add the test for the case that the callback is `partial` and `*arg` and `**kwargs` are also provided by the user

squirrel/iterstream/base.py
```diff
@@ -75,6 +77,7 @@ def async_map(
         buffer: int = 100,
         max_workers: t.Optional[int] = None,
         executor: t.Optional[Executor] = None,
+        **kw,
```
AlirezaSohofi, line 80: please add this to the docstring and explain how this is used

AlirezaSohofi (changes requested):
LG, just some minor comments :+1:

docs/advanced/iterstream.rst
```diff
@@ -91,14 +91,13 @@ The only difference is that we wrap the function inside a custom :py:class:`Comp
...
-    starts (e.g. in the multiprocessing context).
-
-    #. When a very complex  stream processing is needed that is hard to achieve with standard methods.
+    * We need to instantiate an expensive resource such as a database connection or a CNN feature extractor only once
+      for the entire stream instead of once per item. In this case it is advisable to instantiate this resource
+      in the ``__iter__`` method of your custom class. One advantage is that the object is only instantiated once
+      iterating over the stream starts. Additionally, if this resource is a non-picklable object, the stream can still
+      be pickled before the iteration starts (e.g. in the multiprocessing context).
+
+    * When a very complex  stream processing is needed that is hard to achieve with standard methods.
```
axkoenig, line 100:
```suggestion
    * When a very complex stream processing is needed that is hard to achieve with standard methods.
```

docs/usage/iterstream.rst
```diff
@@ -38,7 +38,14 @@ iterable as the input:
         print(item)
 
 :code:`map_async()` applies the provided function asynchronously. More on this in the following sections.
-In addition to explicitly iterating over the items, it's also possible to call `collect()` to collect all items in a list, or `join()` to iterate over items without returning anything.
+To pass additional arguments to the mapped function, simply pass them as additional key-word arguments.
```
axkoenig, line 41:
```suggestion
To pass additional arguments to the mapped function, simply pass them as additional key-word arguments to `map`. This works analogously for `async_map` and `dask_map`.
```

squirrel/iterstream/base.py
axkoenig: agreed. note that in our codebase we mostly use `**kwargs` and not `**kw`. I would suggest to keep this consistent and use `**kwargs` here. 
below are the only two instances of `**kw`, I think this could be a good opportunity to change them to `**kwargs`. 

Screen Shot 2022-09-20 at 11 21 34

axkoenig: please explain each argument briefly in a docstring that has the same format as the one in e.g. `async_map`

squirrel/iterstream/base.py
```diff
@@ -102,11 +105,15 @@ def async_map(
...
         """
-        return _AsyncMap(source=self, callback=callback, buffer=buffer, max_workers=max_workers, executor=executor)
+        partial_callback = partial(callback, **kw)
+        return _AsyncMap(
+            source=self, callback=partial_callback, buffer=buffer, max_workers=max_workers, executor=executor
+        )
 
-    def dask_map(self, callback: t.Callable) -> _Iterable:
+    def dask_map(self, callback: t.Callable, **kw) -> _Iterable:
         """Converts each item in the stream into a dask.delayed object by applying the callback to the item"""
```
axkoenig, line 114: please explain each argument briefly in a docstring that has the same format as the one in e.g. `async_map`

axkoenig: also, please add `*args` here

test/test_iterstream/test_stream.py
```diff
@@ -244,9 +251,13 @@ def _add_one(x: int) -> int:
 def test_dask(samples: t.List[SampleType]) -> None:
```
axkoenig, line 251: I would expect `test_dask` to test `dask_map`. but you're testing `async_map` here. would also test `dask_map` (with args and kwargs) here or do that at in the function `test_different_maps` where `dask_map` is already tested

axkoenig (changes requested):
thanks good work! please see comments, afterwards lgtm

squirrel/iterstream/base.py
pzdkn: adding `*args` might encourage some errors. For example you have 

```
def divide(value, divisor):
        return value / divisor
```
Then for,

`IterableSource(range(1, 3)).map(divide,  2)`

I personally would expect [1/2, 3/2], but gives you [2/1,  2/2]

pzdkn: @AlirezaSohofi @axkoenig, wdyt ?

AlirezaSohofi: @pzdkn Yes you are right, very good catch. I think there is a way to circumvent this but that probably requires refactoring of some of the internal machineries which is out of the scope for this pr. I'd suggest to just add `**kwargs` for now and tackle that in another pr.

axkoenig: sounds good let's just go for `**kwargs` then :)

squirrel/iterstream/base.py
```diff
@@ -99,14 +102,28 @@ def async_map(
...
-    def dask_map(self, callback: t.Callable) -> _Iterable:
-        """Converts each item in the stream into a dask.delayed object by applying the callback to the item"""
-        return self.to(dask_delayed_, callback)
+    def dask_map(self, callback: t.Callable, **kw) -> _Iterable:
+        """
+        Converts each item in the stream into a dask.delayed object by applying the callback to the item.
+        Specify additional keyword arguments via kw
+        Args:
+            callback: callback to be mapped over
+            **kw: key-word arguments for callback
```
axkoenig, line 120: You docstrings don't adhere to the official google docstring style that we're using in squirrel. e.g. you're not adding the type of the argument after the argument name. Same holds for the return statement. you would write `_Iterable: the mapped Squirrel iterable` or sth like that. 
I recommend to simply install a plugin in your IDE that formats this and gives you a skeleton in which you can fill it.

axkoenig (approved):
fine now thanks. only a minor docs update - I think these things are important to improve the docs quality of squirrel. also I would prefer `kwargs` instead of the `kw` which we typically don't use in squirrel - see my comment :)

squirrel/iterstream/base.py
pzdkn: thanks for the hint!

test/test_iterstream/test_stream.py
```diff
@@ -118,13 +124,22 @@ def __iter__(self):
...
     )
-    assert all(i == 4 for i in res)
+    assert all(i == 4 for i in res_1)
+    res_2 = IterableSource(res_1).async_map(lambda value, offset: value + offset, offset=1).collect()
+    assert all(i == 5 for i in res_2)
+    res_3 = IterableSource(res_2).async_map(multiply, factor=2).collect()
+    assert all(i == 10 for i in res_3)
+    linear_fn = partial(lambda x, a, b: a * x + b, b=1)
+    res_4 = IterableSource(res_3).async_map(linear_fn, a=2).collect()
+    assert all(i == 21 for i in res_4)
```
AlirezaSohofi, line 140: minor: it would be easier to check the correctness of the logic if the input is e.g. `list(range(5))` and the assert statement compare the exact list than we expect to see rather than the sum `[1, 2, 3] == [3, ..., ....]`

AlirezaSohofi (dismissed):
LGTM :+1:

test/test_iterstream/test_stream.py
pzdkn: Just fixed this and need another review @AlirezaSohofi

AlirezaSohofi (approved):

PR #92, opened 2022-10-05
zip_index method for Composable

AlirezaSohofi:
# Description

This pr would resolve the issue that shard keys are not sortable. 

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:


squirrel/iterstream/base.py
```diff
@@ -211,6 +211,15 @@ def loop(self, n: t.Optional[int] = None) -> Composable:
         """
         return _LoopIterable(self, n)
 
+    def zip_index(self, pad_length: int = None) -> Composable:
+        """Zip the item in the stream with its index and yield Tuple[index, item]
```
pzdkn, line 215: Setting pad_length to None is the most "dangerous" option actually if the user expects the shards to be sorted.

We should warn/add to the docstring what will happen when the pad_length < length of longest sequences of digits

AlirezaSohofi: This method is generic on its own right, not specifically meant to solve the key ordering, so the docstring does not need to explain that. I'd say a section in the doc or a tutorial notebook is a better place for that.

requirements.txt
```diff
@@ -545,6 +545,9 @@ isodate==0.6.1 \
 itsdangerous==2.1.2 \
     --hash=sha256:2c2349112351b88699d8d4b6b075022c0808887cb7ad10069318a8b0bc88db44 \
     --hash=sha256:5dbbc68b317e5e42f327f9021763545dc3fc3bfe22e6deb96aaf1fc38874156a
+jeepney==0.8.0 \
```
pzdkn, line 548: I guess this should follow a change in the `requirements.in`

pzdkn (approved):
LGTM!

requirements.txt
AlirezaSohofi: This is not our direct dependency

PR #93, opened 2022-10-06
Add csv driver option to specify csv read args

MaxSchambach:
# Description

Adds a `read_csv_kwargs` argument to `CsvDriver` initialization which is used in all `read_csv` calls in the class.
Does not break backward compatibility, as `get_df` and `get_iter` still allow to specify kwargs for `read_csv` which will take precedence over the ones given at initialization.

This makes the creation of new catalog entries based on the `CsvDriver` much easier as dataset-specific read options (such as seperator, dtypes, etc.) can be specified in the `driver_kwargs`.

## Type of change

- [x] Documentation update

# Checklist:


requirements.txt
```diff
@@ -462,62 +462,6 @@ google-resumable-media==2.3.3 \
 googleapis-common-protos==1.56.4 \
     --hash=sha256:8eb2cbc91b69feaf23e32452a7ae60e791e09967d81d4fcc7fc388182d1bd394 \
     --hash=sha256:c25873c47279387cfdcbdafa36149887901d36202cb645a0e4f29686bf6e4417
-greenlet==1.1.2 \
```
pzdkn, line 465: doesn't a change here mean that `requirements.in` also got changed?

MaxSchambach: Dang, I didn't do any requirements changes and did not commit that on purpose. Will revert.

MaxSchambach: Done, thanks for the catch!

MaxSchambach:
> I don't understand, why the test fails due to changes introduced by `black`. I did run `black` locally before submitting and don't see why it fails in the cloud test. Any ideas @pzdkn ?

should be solved now. weirdly, `pre-commit` performed a different `black` call than calling `black` directly.

pzdkn (approved):
Thanks Max, LGTM!

PR #95, opened 2022-10-18
PoC to cache data

AlirezaSohofi:
```
driver = MessagepackDriver(url=url, cache_url=another_url)
```

# Description

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:


squirrel/store/squirrel_store.py
```diff
@@ -70,3 +71,62 @@ def keys(self, nested: bool = False, **kwargs) -> t.Iterator[str]:
...
+        self._store = SquirrelStore(url, serializer)
+
+    def get(self, key: str, **kwargs) -> t.Iterator[SampleType]:
+        """
+        If the item is cached, read from cache, otherwise read from the original source, cache it and stream the items
+        from the shard
+
+        """
+        if not self._is_cached(key):
+            shard = list(self._store.get(key))
```
pzdkn, line 106: Should this `get` call also get kwargs passed ?

squirrel/store/squirrel_store.py
```diff
@@ -70,3 +71,62 @@ def keys(self, nested: bool = False, **kwargs) -> t.Iterator[str]:
...
+            yield from shard
+        else:
+            yield from self._cache.get(key, **kwargs)
+
+    def _is_cached(self, key: str) -> bool:
+        p = f"{self._cache.url}/{key}.gz"
+        return self._cache.fs.exists(p)
+
+
+class CacheStoreSource(Composable):
```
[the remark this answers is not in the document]
AlirezaSohofi: It can and probably it should. We had a chat about caching at the IterStream level with @MaxSchambach and this is the best I could think of, which is not a very sensible abstraction as you pointed out. wdyt @MaxSchambach? Is this useful in your opinion for the kind of situation you anticipate? Otherwise I am happy to remove it.

squirrel/driver/store.py
```diff
@@ -20,20 +21,27 @@ class StoreDriver(MapDriver):
     name = "store_driver"
 
     def __init__(
-        self, url: str, serializer: SquirrelSerializer, storage_options: dict[str, Any] | None = None, **kwargs
+        self,
+        url: str,
+        serializer: SquirrelSerializer,
+        cache_url: str = None,
```
MaxSchambach, line 27: Type annotation of `cache_url` should be `str | None` or `Optional[str]` I think

squirrel/driver/store.py
```diff
@@ -87,5 +95,10 @@ def keys(self, **kwargs) -> Iterable:
     def store(self) -> AbstractStore:
         """Store that is used by the driver."""
         if self._store is None:
-            self._store = SquirrelStore(url=self.url, serializer=self.serializer, **self.storage_options)
+            if self.cache_url is None:
+                self._store = SquirrelStore(url=self.url, serializer=self.serializer, **self.storage_options)
+            else:
+                self._store = CacheStore(
+                    url=self.url, serializer=self.serializer, cache_url=self.cache_url, cache_serializer=self.serializer
```
MaxSchambach, line 102: Since `CashStore` can deal with `storage_options`, would it make sense to add a `cash_storage_options` argument and pass it to the init here? Not sure if that could be useful (definitely not in the standard case of using a local url for cashing) but the functionality is more or less there.

MaxSchambach (approved):
I've added two minor comments, otherwise: looks good to me, thank you!

pzdkn (approved):
LGTM!

PR #102, opened 2022-11-09
Bugfix deserializer kwargs

mg515:
# Description

Currently there is no way to pass `kwargs` to the de-serializer. This can be an issue, especially if being able to set those arguments prevents you to read data samples.
This PR exposes those `kwargs` explicitly to the high level API, as in the `get_iter(loader_kwargs)`

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:


squirrel/serialization/jsonl.py
```diff
@@ -109,7 +109,7 @@ def serialize_shard_to_file(
 
     @staticmethod
     def deserialize_shard_from_file(
-        fp: str, fs: t.Optional[AbstractFileSystem] = None, mode: str = "rb", **open_kwargs
+        fp: str, fs: t.Optional[AbstractFileSystem] = None, mode: str = "rb", open_kwargs: t.Optional[t.Dict] = None,
```
AlirezaSohofi, line 112:
```suggestion
        fp: str, fs: t.Optional[AbstractFileSystem] = None, mode: str = "rb", loader_kwargs: dict[str, Any] | None = None, **open_kwargs
```

AlirezaSohofi (commented):
`loader_kwargs` should be added to `get_iter()` and be passed here

squirrel/driver/jsonl.py
```diff
@@ -38,20 +38,14 @@ def __init__(
 
     def get_iter(
         self,
-        get_kwargs: t.Optional[t.Dict] = None,
```
mg515, line 41: @AlirezaSohofi No more explicit `get_kwargs`, now one either needs to pass `open_kwargs`, `loader_kwargs` or `unpacker_kwargs` to `get_iter(**kwargs)`. How's that from the UX perspective?

pzdkn: I also think it's better to be more explicit. Will we introduce breaking changes by including`open_kwargs` and `loader_kwargs` in `MapDriver.get_iter()` and e.g `JsonDriver` though? I don't think so actually.

mg515: @AlirezaSohofi @pzdkn Apologies for delay on this one. I've adapted the solution to your proposal, as now the `loader_kwargs` are added explicitly to the high level `get_iter()` API and then passed onwards to the `.get()` method. 

The only consideration from my side atm, is that we only expose `loader_kwargs`, even though they are later on used in `Unpacker` (msgpack context) as  well as `.load()` (jsonl context).
We could either find an alternative name that fits both use cases, or expose both kwargs separately (`unpacker_kwargs` as well as `loader_kwargs`). This to me feels a bit overboard. I felt that `loader_kwargs` is a suiting name for both contexts, but please let me know if you have a better idea.

squirrel/driver/driver.py
```diff
@@ -73,6 +72,7 @@ def get_iter(
         flatten: bool = False,
         keys_kwargs: dict | None = None,
         get_kwargs: dict | None = None,
+        loader_kwargs: dict | None = None,
```
pzdkn, line 75: Can you update the doc string with loader_kwargs pls ?

squirrel/driver/jsonl.py
```diff
@@ -42,19 +42,24 @@ def __init__(
...
+        loader_kwargs: t.Optional[t.Dict] = None,
         **kwargs,
     ) -> Composable:
         """Returns an iterable of samples as specified by `fetcher_func`.
 
         Args:
             get_kwargs (Dict): Keyword arguments that will be passed as `get_kwargs` to :py:meth:`MapDriver.get_iter`.
                 `get_kwargs` will always have `compression="gzip"`. Defaults to None.
+            loader_kwargs (Dict): Keyword arguments that will be passed
+                as `loader_kwargs` to :py:meth:`MapDriver.get_iter`.
```
pzdkn, line 54: might be more clear to say that it gets passed to `jsonl.load` directly

pzdkn (changes requested):
Thanks Miha, looks great. Just need some documentation updates.

PR #104, opened 2022-12-19
Interaction Nvidia DALI and Squirrel 

axkoenig:
# Description

Describes in detail how Squirrel and DALI can work together. Also includes benchmarks on how to best utilize DALI and how it compares to transforms in Torchvision.

Attaching PDF rendered version of the Sphinx documentation [here](https://github.com/merantix-momentum/squirrel-core/files/10260560/NVIDIA.DALI.Squirrel.documentation.pdf). Unfortunately, I couldn't get syntax highlighting to work.  

Apparent next steps are figuring out how Squirrel and DALI can work together in multi-processing. It is not obvious how we could implement this, and if this provides a performance boost. Using a DALI [parallel external source](https://docs.nvidia.com/deeplearning/dali/user-guide/docs/examples/general/data_loading/parallel_external_source.html) would probably be the way to go, but DALI expects a callable here that fetches individual images given a specific image index. This can be implemented easily if we set shard-size=1, but our initial experiments showed that larger shard sizes are more desirable. 

## Type of change

- [x] Documentation update

# Checklist:


pzdkn (approved):
Nice work! LGTM! 

Its surprising that increasing threads doesn't have much impact. Do you have an intuition on this ?

PR #105, opened 2023-01-17
Support running all tests in python3.7

winfried-ripken:
# Description

I checked for compatibility with python 3.7 - all the code runs ok, but still some tests failed. I fixed some minor issues s.t. all tests run successfully in py3.9 now.

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/test/test_driver/test_msgpack.py
+++ b/test/test_driver/test_msgpack.py
@@ -1,3 +1,4 @@
+import shutil
 import tempfile
 
 import torch.utils.data as tud
@@ -5,7 +6,7 @@ import torch.utils.data as tud
 from squirrel.constants import URL
 from squirrel.driver import MessagepackDriver
 from squirrel.iterstream import IterableSource
-from squirrel.iterstream.torch_composables import TorchIterable, SplitByWorker
+from squirrel.iterstream.torch_composables import SplitByWorker, TorchIterable
 from squirrel.serialization import MessagepackSerializer
 from squirrel.store import SquirrelStore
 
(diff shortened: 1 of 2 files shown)


test/test_driver/test_msgpack.py
```diff
@@ -36,13 +37,18 @@ def test_get_iter(dummy_msgpack_store: SquirrelStore, num_samples: int) -> None:
 
 def test_clean_store() -> None:
     """Test instantiating a store and removing all it's content"""
-    with tempfile.TemporaryDirectory() as tmp_dir:
```
AlirezaSohofi, line 39: Is this test causing issue wrt python version? 
It's generally better to use context manager for these cases. Best would be to use a fixture (see [example](https://github.com/merantix-momentum/squirrel-core/blob/main/test/conftest.py#L159)).

AlirezaSohofi (changes requested):

test/test_driver/test_msgpack.py
winfried-ripken: yes, the problem is that the directory is already cleaned up by squirrel as expected - this is case is however not checked for in python v37 within the tempfile.TemporaryDirectory() context and therefore the test fails with an exception. Happy to discuss alternatives or leave it as is - this would mean that tests fail on py37 but the core functionality should still work ok

AlirezaSohofi: :+1:

AlirezaSohofi (approved):

PR #109, opened 2023-01-30
sliding window on stream with stride

AlirezaSohofi:
# Description

This PR provides a way to generate sliding windows from a stream of items. It's useful for e.g. training autoregressive models and fixed-window aggregation. It is a generalization of `Composable.batched()`: when `window_size == stride`, the two are equivalent. 

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:


squirrel/iterstream/base.py
```diff
@@ -173,6 +173,40 @@ def batched(
...
+
+        Args:
+            window_size (int): the length of the window
+            deepcopy (bool): If True, each window will be returned as a deepcopy. If items are mutated in the
+                subsequent steps of the pipeline, this should be set to True, otherwise it should be False.
+                Note that deepcopy may incur a substantial cost, so set this parameter carefully.
+            stride (int): the distance that the window moves at each step
+            drop_last_if_not_full (bool): If True, it would only return windows of size `window_size` and drops the
+                last items which have fewer items.
+            min_window_size (int): The minimum length of the window. This argument is only relevant if
```
SebS94, line 196: Is this the common way of handling incomplete windows? As a user the more intuitive behaviour for me would be to supply  a default value which is used to complete the last window - then I can filter/ignore it in further processing steps as necessary?

SebS94 (approved):
LGTM - except for the comment on the expected behaviour.

squirrel/iterstream/base.py
AlirezaSohofi: make it possible to fill partial windows with None is a good suggestion, will add this, thanks :+1:

squirrel/iterstream/base.py
```diff
@@ -173,6 +173,40 @@ def batched(
...
+                `drop_last_if_not_full` is set to False, otherwise it's ignored.
```
pzdkn, line 197: It is not immediately clear to me  what `min_window_size`  is the minimum window size for. One could add that it is the minimum window size for the last remaining elements.

squirrel/iterstream/base.py
```diff
@@ -317,6 +351,59 @@ def __iter__(self) -> t.Iterator:
...
+    def __iter__(self):
+        it = iter(self.source)
+        _win = []
+
+        while len(_win) < self.window_size:
+            try:
+                _win.append(next(it))
+            except StopIteration:
+                if not self.drop_last_if_not_full:
+                    yield _win
```
pzdkn, line 381: I guess we need to yield a deep-copied version here too

AlirezaSohofi: nice catch! Thank you :+1:

squirrel/iterstream/base.py
pzdkn: Or even changing name to `min_remaining_items`

pzdkn (dismissed):
LGTM! Thanks Ali!

pzdkn (approved):
LGTM! Thanks!

pzdkn (approved):
LGTM!

PR #113, opened 2023-02-23
local_cache_dir argument for DataFrameDriver

AlirezaSohofi:
# Description

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/squirrel/driver/csv.py
+++ b/squirrel/driver/csv.py
@@ -35,4 +35,10 @@ class CsvDriver(DataFrameDriver):
         else:
             import pandas as pd
 
-            return pd.read_csv(self.url, **kwargs)
+            if self.local_caching:
+                try:
+                    return pd.read_csv(self.local_caching, **kwargs)
+                except FileNotFoundError:
+                    df = pd.read_csv(self.url, **kwargs)
+                    df.to_csv(self.local_caching, index=False)
+                    return df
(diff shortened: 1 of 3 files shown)


ditschuk (commented):
I commented on a problem I see with the caching based on `pd.read_csv`

MaxSchambach:
Thanks for this quick draft! I think it can pretty much be as easy, but would suggest to move the caching functionality to the parent class `DataFrameDriver` so that all data frame drivers (csv, excel, feather, parquet, ...) benefit from it such that basically the derived classes do not need any changes. Including Konstantin's valid comment, `DataFrameDriver` could deal with downloading the file using fspec when `local_cache_dir` is not `None` and setting `self.uri` correspondingly and the derived classes would need no changes.

MaxSchambach (changes requested):
Thanks for this quick draft! I think it can pretty much be as easy, but would suggest to move the caching functionality to the parent class `DataFrameDriver` so that all data frame drivers (csv, excel, feather, parquet, ...) benefit from it such that basically the derived classes do not need any changes. Including Konstantin's valid comment, `DataFrameDriver` could deal with downloading the file using fspec when `local_cache_dir` is not `None` and setting `self.uri` correspondingly in `get_df` and the derived classes would need no changes as they merely implement `_read()`.

PR #117, opened 2023-03-31
Add name to dataframe driver

pzdkn:
# Description

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/squirrel/driver/data_frame.py
+++ b/squirrel/driver/data_frame.py
@@ -20,6 +20,9 @@ ENGINE = Literal["dask", "pandas"]
 
 
 class DataFrameDriver(FileDriver, metaclass=ABCMeta):
+
+    name = "dataframe"
+
     def __init__(
         self,
         url: URL,


MaxSchambach (commented):
Since `DataFrameDriver` is an abstract base class that cannot be instantiated, I am hesitant to add the `name` attribute. Similarly, the base `Driver` class as well as `IterDriver` etc. do not have a `name` attribute. From a user's perspective, the name attribute might suggest that the driver can directly be used, e.g. in the definition of a `Source` which is not the case. Therefore, I would opt for not adding the attribute.

@pzdkn what is your intuition/reason for adding this? Maybe I misunderstand the purpose.

AlirezaSohofi (approved):
LG :+1:

AlirezaSohofi:
> Since `DataFrameDriver` is an abstract base class that cannot be instantiated, I am hesitant to add the `name` attribute. Similarly, the base `Driver` class as well as `IterDriver` etc. do not have a `name` attribute. From a user's perspective, the name attribute might suggest that the driver can directly be used, e.g. in the definition of a `Source` which is not the case. Therefore, I would opt for not adding the attribute.
> 
> @pzdkn what is your intuition/reason for adding this? Maybe I misunderstand the purpose.

You are right. In terms of functionality it does not matter. If you think it adds confusion rather than clarity we can leave it out.

MaxSchambach:
Yes, since I think it does not add any functionality, as the class cannot be instantiated, and has the potential to create confusion, I would opt for leaving it out.

PR #121, opened 2023-05-12
Enable Caching and Docs Update

axkoenig:
# Description

This is a follow up for our PRs #113 and PRs #95. 

Since fsspec comes with an advanced caching functionality, we do not need to have a `CacheStore` for our `StoreDriver`. Rather we should use the powerful API provided by fsspec, which ensures a unified caching mechanism across **all** of our drivers.

The PR
- removes the old caching functionality and its tests
- implements minor changes in filesystem initialisation, such that `protocol` can be set through `storage_options`
- implements a minor change in how `storage_options` can be updated via `CatalogSource.get_driver()`. The default behavior was that any newly provided `storage_options` in `get_driver` would entirely overwrite the existing `storage_options` of the `CatalogSource`, which leads to a loss of potentially important configs of the person who gave you the catalog (e.g. some bucket configs like `requester_pays` should not be overwritten!). 
- a test on this ^
- documentation update on how to use fsspec caching with squirrel, some context around the topic and some minor benchmarks

## Type of change

- [x] Documentation update

# Checklist:


test/test_catalog/test_catalog.py
```diff
@@ -179,6 +178,30 @@ def test_catalog_repr() -> None:
...
+    }
+    so_rp = {"requester_pays": True}
+
+    # create source and register in catalog
+    cat = Catalog()
+    cat["source"] = Source("csv", driver_kwargs={"url": "gs://some-bucket/test.csv", "storage_options": so_rp})
+
+    # update storage_options via get_driver
+    driver = cat["source"].get_driver(storage_options=so_cache)
+    so_all = {**so_rp, **so_cache}
```
MaxSchambach, line 201: For testing, please make this explicit, as the same (rather implicit) overwriting of keywords is used in the implementation. I.e., provide a ground truth dict with all expected key-value pairs.

test/test_catalog/test_catalog.py
```diff
@@ -214,7 +237,7 @@ class TestDriver(Driver):
         name = "testdriver"
 
         def __init__(self, welcome: str, **kwargs) -> None:
-            super().__init__(*kwargs)
+            super().__init__(**kwargs)
```
MaxSchambach, line 240: How did that work before? 🙃

test/test_catalog/test_catalog.py
MaxSchambach: Also please add assert without storage option update and add a keyword besides `protocol` that is overwritten. I.e. something like 
```
# Use default storage_options
driver = cat["source"].get_driver()
assert ...

# update storage_options via get_driver
driver = cat["source"].get_driver(storage_options=so_cache)
so_all = {<key>: <value>, ...}
assert ...
```

MaxSchambach (changes requested):
Thanks, I think this already looks good! I've left some minor remarks.
Also, since these are breaking changes regarding the `StoreDriver` and some low-level changes in how fsspec options are handled, I think @AlirezaSohofi should have a quick look as well and give his approval.

docs/advanced/caching.rst
```diff
@@ -0,0 +1,77 @@
...
+among other things. For more detailed information, please refer to the ``fsspec``
+`documentation <https://filesystem-spec.readthedocs.io/en/latest/features.html#caching-files-locally>`_
+on local caching.
+
+The code below shows an example of configuring caching for several drivers. Note that, as per the
+``fsspec`` documentation, only ``simplecache`` is "guaranteed thread/process-safe".
+
+.. code-block:: python
+
+    from squirrel.driver import CsvDriver, FileDriver, MessagepackDriver
```
pzdkn, line 43: maybe mention some other caching options and how to choose them. tbh I couldn't find any on gcfs and fsspec docs

pzdkn: it seems they are also target_protocol dependent, is there a reference we can add here ?

squirrel/catalog/catalog.py
```diff
@@ -357,6 +357,12 @@ def get_driver(self, **kwargs) -> Driver:
         for plugin in plugins:
             for driver_cls in plugin:
                 if driver_cls.name == self.driver_name:
+                    # below line ensures that providing storage_options in the kwargs does not overwrite
+                    # the present storage_options, but rather updates them with the newly provided ones.
+                    kwargs["storage_options"] = {
+                        **self.driver_kwargs.get("storage_options", {}),
+                        **kwargs.get("storage_options", {}),
+                    }
                     return driver_cls(catalog=self._catalog, **{**self.driver_kwargs, **kwargs})
```
pzdkn, line 366: I think line.366 updates the dict anyway, so line.360 is redundant. However, we might want to think about how to make it more obvious what is happening here.,

docs/advanced/caching.rst
```diff
@@ -0,0 +1,77 @@
...
+
```
pzdkn, line 44: any gotchas here? what if the shard name stays the same but the shard changes? what is the caching mechanism based on (name or contents?)

pzdkn (commented):
Thanks Alex, looks great! I left some comments here.

squirrel/fsspec/fs.py
```diff
@@ -31,9 +31,12 @@ def get_fs_from_url(url: URL, **storage_options) -> FILESYSTEM:
     GCSFileSystem.` I suggest you instead of passing `requester_pays=True` to `storage_options` in fsspec, simply
     switch to the right project by `gcloud config set project PROJECT_ID` where you have the right role(s). The problem
     should be resolved by itself.
+
+    The protocol can be overriden via the `storage_options`. This is important if people want to use the fsspec caching
+    functionality, which requires the protocol to be, e.g., `simplecache`.
     """
     protocol, _ = split_protocol(url)
-    return fsspec.filesystem(protocol, **storage_options)
+    return fsspec.filesystem(**{"protocol": protocol, **storage_options})
```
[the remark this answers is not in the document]
axkoenig: thanks I integrated that

test/test_catalog/test_catalog.py
axkoenig: I added explicit checking 
I added check without storage option update
I'm not sure I understand your last point. There are already mutliple keywords besondes `protocol` in storage options that are overwritten. Or do you mean I should overwrite another keyword of the driver, like overwriting the url of the driver? I have implemented a check if the "usual way of overwriting kwargs" still works - I assume you meant that. please double check the code once I pushed :)

test/test_driver/test_cache.py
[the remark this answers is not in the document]
axkoenig: we could have some small dataset on our buckets or just download some file from the internet using filedriver? e.g. the squirrel logo from github or something like that. that should work no? 
in terms of URI I found this [here](https://github.com/fsspec/filesystem_spec/blob/a85876a6f8a0751adccbeed9defc12e8e69f59f9/fsspec/tests/test_caches.py#L119) which looks like it could be interesting to us, but I can't find what the variable "server" is set to, possibly that's hidden from the public somewhere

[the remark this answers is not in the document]
axkoenig: I think @AlirezaSohofi  could have an idea here / decide on whether we can / should test fsspec again

squirrel/fsspec/fs.py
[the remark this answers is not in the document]
axkoenig: as @MaxSchambach mentioned, we would need your input here @AlirezaSohofi . do you see any possible problem that could be introduced by this lower-level change? it was necessary to correctly pass down the storage option arguments which enable the caching mechanism

squirrel/driver/store.py
```diff
@@ -24,28 +23,21 @@ def __init__(
...
-        cash_storage_options: dict[str, Any] | None = None,
         **kwargs,
     ) -> None:
         """Initializes StoreDriver.
 
         Args:
             url (str): the url of the store
             serializer (SquirrelSerializer): serializer to be passed to SquirrelStore
-            cache_url (str): if provided, the data will be cached in a store at this url
             storage_options (Optional[Dict[str, Any]]): a dict with keyword arguments to be passed to store initializer
```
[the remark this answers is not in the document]
axkoenig: ok I added a note to all drivers resurfacing this info

docs/advanced/caching.rst
axkoenig: there's a pretty extensive docs on caching options at fsspec [here](https://filesystem-spec.readthedocs.io/en/latest/features.html#caching-files-locally)
I would suggest not to replicate their explanations, but I rather linked their documentation

docs/advanced/caching.rst
axkoenig: but I haven't worked much with this beyond this implementation so can't really add anymore gotchas / learnings

squirrel/catalog/catalog.py
axkoenig: I'm unsure what you mean here, line 360 in my code is just a comment. I tried to explain my code additions in the two comments in line 360 and 361. it just makes sure that the original storage_options of the catalog source don't get lost when some incremental updates to the storage_options are passed down when instantiating a driver through the catalog API.

axkoenig:
Hi folks, thanks for your reviews. I addressed all your comments and just pushed the code, please have another look and maybe reply to my comments if you feel there's anything to discuss. We need some final review from @AlirezaSohofi mainly to check out the minor lower-level change of how the storage options are passed down in `squirrel/fsspec/fs.py`
Cheers, Alex

squirrel/catalog/catalog.py
MaxSchambach: I agree that it's a bit tricky, what's happening here. In 360 and following, the storage options are updated using the given `kwargs["storage_options"]`. In 366, the driver kwargs are again updated. Basically, one needs 360 and following to recursively update the nested kwargs + storage options. As the comment explains, otherwise the full `storage_options` dict would be replaced instead of updated. Maybe the comment could be extended to clarify this even more.

MaxSchambach (dismissed):
Thanks for the update! WIth one or two minor modifications, LGTM! Thanks for pushing this over the line @axkoenig 🙌

AlirezaSohofi (changes requested):

docs/advanced/caching.rst
```diff
@@ -0,0 +1,75 @@
...
+    so = {"protocol": "simplecache", "target_protocol": "gs", "cache_storage": "/tmp/cache"}
```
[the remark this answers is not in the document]
axkoenig: ok I changed all instances to `path/to/cache`. I did not use `/path/to/cache` un purpose to not spam the user's root directory if they execute any of the example code

squirrel/catalog/catalog.py
axkoenig: Ok I updated the comment to give a more comprehensive description

squirrel/fsspec/fs.py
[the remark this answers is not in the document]
axkoenig: resolving since no comment/warning from Ali

test/test_catalog/test_catalog.py
```diff
@@ -179,6 +178,42 @@ def test_catalog_repr() -> None:
     assert cat.__repr__() == "['a', 'ab', 'b', 'c']"
 
 
+def test_get_driver_storage_options() -> None:
```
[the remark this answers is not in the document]
axkoenig: Hi I added a test for the get_fs_from_url function
It seems like a good idea to have a brief real-world test where there is data streamed and cached from a remote bucket. how do you suggest to proceed with this? Which bucket is public and could be used for this? In the existing tests I could not find a remote repository with sample data for tests

test/test_catalog/test_catalog.py
axkoenig: regarding your first, point this is exactly what I do [here](https://github.com/merantix-momentum/squirrel-core/blob/9cd0625a1d8b5f38b2dc1353ef05a6cc32ffd8ec/test/test_catalog/test_catalog.py#L230). For some reason I think the github UI shows an outdated version of the code here.. Instead of `== so_rp` I write `== {"requester_pays": True}`, which should be more explicit like you said. 
ok, I added another check for updating existing storage_options :)

axkoenig: I am now resolving this, please feel free to reopen if anything is missing

test/test_driver/test_cache.py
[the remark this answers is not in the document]
axkoenig: closing this thread as ali also raised this concern and I will address it down below

test/test_catalog/test_catalog.py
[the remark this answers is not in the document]
axkoenig: Hi there,
thank you so much for the hint. Using this bucket it was easy to implement the caching tests [here](https://github.com/merantix-momentum/squirrel-core/blob/c2d7690040eb793d53561d464040d9458d102e4b/test/test_driver/test_caching.py#L26). I implemented a test for the CSV driver for now (since this inherits from DataframeDriver, I view this as a sufficient test for the DataframeDriver parent class). Or do you suggest to run through tests for every single class that inherits from DataframeDriver (e.g. I would also test ParquetDriver?). 
Similarly, I would implement caching tests for MessagepackDriver (which should cover MapDriver-style drivers) and a FileDriver caching test, since this is a special one. Otherwise I can also implement caching tests for all available drivers like JsonL, Feather, Excel, ... if you like - but this comes at an increased maintenance cost in the future, as we'll have to remember to implement tests for these drivers in the future too. WDYT @AlirezaSohofi?

axkoenig:
@AlirezaSohofi the branch trigger above fails because of `ModuleNotFoundError: No module named 'numba' `, do you have a quick fix for this or any hints?

test/test_catalog/test_catalog.py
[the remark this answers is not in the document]
AlirezaSohofi: I think testing `CSVDriver` and `MessegepackDriver` is sufficent.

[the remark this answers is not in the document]
axkoenig: now implemented :)

test/test_driver/test_caching.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def test_caching_csv(cache_dir: str) -> None:
+    """Tests caching functionality of DataFrameDriver via CsvDriver."""
+
+    # example data
+    df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
+
+    # upload example data to remote bucket
+    remote_path = REMOTE_DIR + "/test.csv"
```
AlirezaSohofi, line 26: `/test.csv` -> `f"/test_{random.randint(1, 100000)}.csv"`

AlirezaSohofi (dismissed):
LGTM, Thanks so much :+1: 

Left a minor comment. Please don't forget to update the version.

test/test_driver/test_caching.py
axkoenig: sounds great, added that to CSV and msgpack driver tests

AlirezaSohofi (approved):
Thanks @axkoenig :+1:

PR #127, opened 2023-06-20
Fix composable loop reinit bug

MaxSchambach:
# Description
Fixes a tiny but significant bug in `loop` method of `Composable` class.
Running an infinite loop previously resulted in yielding only the very first item of the IterableSource due to a reinitialization of the iterator caused by a wrong indentation level.

The fix is straightforward by correcting the indentation level of the iterator reinitialization.

A test is added to testcase the infinite-loop setting which was previously untested.

See the test or this simple snippet to showcase the bug:

```python

source = IterableSource(range(3)).loop()

for i, x in enumerate(source):
    print(x)
    if i == 9:
        break
# previously yielded [0, 0, ..., 0], should yield [0, 1, 2, ..., 1, 2]
```

## Type of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "squirrel-core"
-version = "0.19.0"
+version = "0.19.1"
 description = "Squirrel is a Python library that enables ML teams to share, load, and transform data in a collaborative, flexible and efficient way."
 authors = ["Merantix Momentum"]
 license = "Apache 2.0"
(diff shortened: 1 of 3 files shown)


AlirezaSohofi (approved):
Nice catch, thanks a lot for the fix :+1:

AlirezaSohofi:
@MaxSchambach can you please increment the version?

MaxSchambach:
Thanks! Version is bumped.

PR #128, opened 2023-06-28
wip: define itertuples kwargs on the initialization of the driver

AlirezaSohofi:
# Description

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:


DominiquePaul:
@AlirezaSohofi I added the mapping function and wrote tests for all configurations. Let me know if this is good to merge :)

squirrel/driver/data_frame.py
```diff
@@ -27,6 +27,8 @@ def __init__(
         engine: ENGINE = "pandas",
         df_hooks: Iterable[Callable] | None = None,
         read_kwargs: dict | None = None,
+        itertuples_kwargs: dict | None = None,
```
MaxSchambach, line 30: Please add a description for the two new options to the docstring.

squirrel/driver/data_frame.py
```diff
@@ -98,5 +102,11 @@ def get_iter(self, itertuples_kwargs: dict | None = None, read_kwargs: dict | No
             (squirrel.iterstream.Composable) Iterable over the rows of the data frame as namedtuples.
         """
         itertuples_kwargs = itertuples_kwargs or {}
+        itertuples_kwargs = {**itertuples_kwargs, **self.itertuples_kwargs}
```
MaxSchambach, line 105: This should probably be the other way around, so that the given `itertuples_kwargs` take precedence over `self.itertuple_kwargs`, i.e.

```python
itertuples_kwargs = {**self.itertuples_kwargs, **itertuples_kwargs}
```

squirrel/driver/data_frame.py
```diff
@@ -98,5 +102,11 @@ def get_iter(self, itertuples_kwargs: dict | None = None, read_kwargs: dict | No
...
         read_kwargs = read_kwargs or {}
-        return IterableSource(self.get_df(**read_kwargs).itertuples(**itertuples_kwargs))
+        it = IterableSource(self.get_df(**read_kwargs).itertuples(**itertuples_kwargs))
+        if self.convert_row_to_dict:
+            def row_to_dict(x):
```
MaxSchambach, line 109: please add a type annotation and docstring

test/test_driver/test_data_frame.py
```diff
@@ -78,8 +78,20 @@ def engine(request: Request) -> ENGINE:
...
 
+@pytest.fixture(params=[True, False])
+def convert_row_to_dict(request: Request) -> bool:
+    """Convert row to dict."""
+    return request.param
+
+
+@pytest.fixture(params=[True, False])
+def itertuples_kwargs(request: Request) -> dict:
+    """temp."""
```
MaxSchambach, line 89: Please update

test/test_driver/test_data_frame.py
```diff
@@ -78,8 +78,20 @@ def engine(request: Request) -> ENGINE:
...
 
 
+@pytest.fixture(params=[True, False])
+def convert_row_to_dict(request: Request) -> bool:
+    """Convert row to dict."""
+    return request.param
+
+
+@pytest.fixture(params=[True, False])
+def itertuples_kwargs(request: Request) -> dict:
```
MaxSchambach, line 88: can be more explicit `dict[str, bool]` (it's a bit overkill here, but hey, good practice :) )

test/test_driver/test_data_frame.py
```diff
@@ -78,8 +78,20 @@ def engine(request: Request) -> ENGINE:
...
+    """temp."""
+    return {"index": request.param}
+
+
 @pytest.fixture
-def data_frame_source(data_frame_source_path: tuple[str, URL, dict], engine: ENGINE) -> tuple[Source, ENGINE]:
+def data_frame_source(data_frame_source_path: tuple[str, URL, dict], engine: ENGINE, convert_row_to_dict: bool, itertuples_kwargs: dict) -> tuple[Source, ENGINE]: #
```
MaxSchambach, line 94: please remove `#` (also in the code below)

test/test_driver/test_data_frame.py
```diff
@@ -105,8 +117,16 @@ def test_dataframe_drivers(data_frame_source: tuple[Source, ENGINE], data_frame_
...
     for (row, series_gt), data in zip(df_gt.iterrows(), driver.get_iter()):
-        d = data._asdict()
-        del d["Index"]
+        if driver.convert_row_to_dict:
+            assert isinstance(data, dict), "Data should be a dict if convert_row_to_dict is True."
+            d = data
+        else: 
+            d = data._asdict()
+        itertuples_index = driver.itertuples_kwargs.get("index", None)
+        if itertuples_index == False:
```
MaxSchambach, line 126: replace with simply `if not itertuples_index`

MaxSchambach (dismissed):
Thank you for the contribution! It looks already good, I just left a few minor comments. Please then also run the pre-commit hook so the test passes

squirrel/driver/data_frame.py
AlirezaSohofi: Both are fine for me, though the current order makes more sense to me. If arguments are explicitly provided at run time, they should be used to overwrite default ones in the catalog. 

In any case, the behavior should be clearly documented.

MaxSchambach: I agree, that arguments provided at runtime should overwrite the default ones, that's my point exactly :)
The current code, however, prioritizes `self.itertuples_kwargs`, e.g. check this mini example
```python
d1 = dict(a=1, b=2)
d2 = dict(b=3, c=4)
print({**d1, **d2}
# >>> {'a': 1, 'b': 3, 'c': 4}
```

AlirezaSohofi: ah right :+1:

DominiquePaul:
Everything should be fixed now @MaxSchambach

AlirezaSohofi:
@DominiquePaul Can you please re-base this branch and make sure that version is correctly incremented?

PR #139, opened 2023-08-16
Add counter to _LoopIterable

MaxSchambach:
# Description

Tiny extension to add a counter to loop iterables holding the current number of iterations of an iterable.
This can be useful to track the current state of that iterable, in particular when using it with infinite loops.
E.g. our use case is that we sample from multiple datasets that each loop indefinitely. Depending on their size, some datasets loop more often than others and it would be nice, e.g. at the end of a training, to see how many times each dataset was iterated over.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "squirrel-core"
-version = "0.19.3"
+version = "0.19.4"
 description = "Squirrel is a Python library that enables ML teams to share, load, and transform data in a collaborative, flexible and efficient way."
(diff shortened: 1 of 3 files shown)


AlirezaSohofi (dismissed):
LG :+1: 
Could you please also update the version?

MaxSchambach:
@AlirezaSohofi I've bumped the version and added a test 👍

AlirezaSohofi (changes requested):
Found a subtle bug

AlirezaSohofi (approved):
LG :+1:

PR #151, opened 2023-10-23
Artifact Manager

SebS94:
# Description

This PR introduces a basic artifact manager functionality based on squirrel stores.

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:


squirrel/artifact_manager/wandb.py
```diff
@@ -0,0 +1,46 @@
+from pathlib import Path
+from typing import Optional, Any, Union, List, Iterable
+
+from squirrel.artifact_manager.base import ArtifactManager
+from squirrel.catalog import Catalog, Source
+
+
+class WandbArtifactManager(ArtifactManager):
```
ThomasWollmann, line 8: please add a comment, that this is a dummy

squirrel/artifact_manager/fs.py
```diff
@@ -0,0 +1,193 @@
...
+}
+
+
+class ArtifactFileStore(FilesystemStore):
+    def complete_key(self, key: str, **open_kwargs) -> List[str]:
+        """Returns a list of possible key continuations given a partial key."""
+        if not self.fs.exists(f"{self.url}/{key}", **open_kwargs):
+            return []
+        return [
+            os.path.relpath(path, f"{self.url}/{key}")
```
ThomasWollmann, line 24: please use pathlib

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,92 @@
...
+    ) -> str:
+        """Get full qualified path or wandb directory of artifact"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def log_file(self, local_path: Path, name: str, collection: Optional[str] = None) -> Source:
+        """Upload file into (current) collection, increment version automatically"""
+
+    @abstractmethod
+    def log_collection(self, files: Union[Path, List[Path]], collection_name: Optional[str] = None) -> Catalog:
```
AlirezaSohofi, line 73: `List[Path]` is generic enough, no need to `Path`

squirrel/artifact_manager/fs.py
```diff
@@ -0,0 +1,193 @@
...
+        """Persists an item with the given key."""
+        if key is None:
+            key = get_random_key()
+        if isinstance(value, Path):
+            self.fs.cp(value, f"{self.url}/{key}/file", **open_kwargs)
+        else:
+            super().set(value, f"{key}/{Serializers[self.serializer.__class__]}", mode, **open_kwargs)
+
+
+class FileSystemArtifactManager(ArtifactManager):
```
AlirezaSohofi, line 51: please mention what directory structure s assumed

test/test_artifact_manager/test_logging.py
```diff
@@ -0,0 +1,56 @@
...
+    tmpdir = tempfile.TemporaryDirectory()
+    manager = FileSystemArtifactManager(url=tmpdir.name, auto_mkdir=True)
+    manager.log_object(obj, artifact_name, collection)
+    assert manager.get_artifact("john", collection) == obj
+    manager.log_object(obj1, artifact_name, collection)
+    assert manager.get_artifact("john", collection) == obj1
+    manager.log_object(obj2, artifact_name, collection)
+    assert manager.get_artifact(artifact_name, collection) == obj2
+    assert manager.get_artifact(artifact_name, collection, 2) == obj1
+    assert manager.get_artifact(artifact_name, collection, 1) == obj
```
AlirezaSohofi, line 56: `manager.get_artifact(artifact_name, collection, 4)`

AlirezaSohofi (commented):
LG in general. I would suggest to add timestamp as well. We can then add search by time functionality.

The `default` collection is a bit odd. I assume that the typical usage will be informally assuming some semantic or schema consistency for each collection and organize artifact by that, then default because like `whatever`. I think there is also a risk that people forget to specify the collection and incorrectly put artifacts in default collection, which can be hard to notice and hard to fix.

I think a nice extension is for artifact registry would be collections with explicit schema definition, say in `json-schema` format. The benefits:
- add validation logic
- amenable to build analytic layers on top
- better search capabilities, e.g. with SQL

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,92 @@
...
+
+class ArtifactManager(ABC):
+    def __init__(self):
+        """
+        Artifact manager interface for various backends
+
+        Maintains a mapping of artifact names to backend objects to facilitate logging and retrieval of arbitrary
+        artifacts.
+        """
+        self._collection = "default"
```
ThomasWollmann, line 16: I assumed that the artifact manager can manage multiple collections. Do I have to maintain multiple artifact manager instances to mange multiple collections? This would seem cumbersome to me.

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,92 @@
...
+        The manager maintains an 'active' collections which it logs to by default.
+        To avoid incompatibility between different backends, collections cannot be nested (e.g. as subfolders on a
+        filesystem).
+        """
+        return self._collection
+
+    @collection.setter
+    def collection(self, value: str) -> None:
+        """Do not allow access to anything beyond the root location of the artifact store"""
+        assert len(value) > 0 and "\\" not in value, "Invalid collection name - must not be empty and cannot be nested."
```
ThomasWollmann, line 32: please use a regex here and a whitelist of characters instead of a blacklist

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,92 @@
...
+    def store_to_catalog(self) -> Catalog:
+        """Provide a catalog of all stored artifacts."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def collection_to_catalog(self, collection: Optional[str] = None) -> Catalog:
+        """Catalog of all artifacts within a specific collection."""
+
+    @abstractmethod
+    def get_artifact(self, artifact: str, collection: Optional[str] = None, version: Optional[int] = None) -> Any:
```
ThomasWollmann, line 50: instead of collection None, use the default collection name?

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,92 @@
...
+        """Retrieve specific artifact value."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_artifact_source(
```
ThomasWollmann, line 55: would not implement artifact logic here and keep cohesion high, but let a user convert this manager to a catalog and let him/her retreive the source from there.

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,92 @@
...
+        self, artifact: str, collection: Optional[str] = None, version: Optional[int] = None
+    ) -> Source:
+        """Catalog entry for a specific artifact"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_artifact_location(
+        self, artifact_name: str, collection: Optional[str] = None, version: Optional[int] = None
+    ) -> str:
+        """Get full qualified path or wandb directory of artifact"""
```
ThomasWollmann, line 65: the manager is not wandb specific

squirrel/artifact_manager/base.py
ThomasWollmann: parameter collection/collection_name not consistent

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,92 @@
...
+        """Upload folder or collection of files"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def log_object(self, obj: Any, name: str, collection: Optional[str] = None) -> Source:
+        """Log an arbitrary python object"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def download_artifact(
```
ThomasWollmann, line 83: use same naming as log_file e.g. download_file

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,92 @@
...
+    def log_file(self, local_path: Path, name: str, collection: Optional[str] = None) -> Source:
+        """Upload file into (current) collection, increment version automatically"""
+
+    @abstractmethod
+    def log_collection(self, files: Union[Path, List[Path]], collection_name: Optional[str] = None) -> Catalog:
+        """Upload folder or collection of files"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def log_object(self, obj: Any, name: str, collection: Optional[str] = None) -> Source:
```
ThomasWollmann, line 78: how to retrieve objects again? can we restrict it to safe object types? e.g. safetensors, numpy, primitive types

test/test_artifact_manager/test_logging.py
SebS94: This would just fail, right?

squirrel/artifact_manager/base.py
SebS94: No - the internal collection is meant to be active so that it can be set once at the start e.g. based on the workflow execution and then all logging calls which to not have an explicit collection will log there.

squirrel/artifact_manager/base.py
SebS94: See above - if no collection is provided, the current active collection is used.

squirrel/artifact_manager/base.py
SebS94: I thought the source description of squirrel would be a nice vehicle to expose for pre-processing and filtering artifacts before fetching/downloading them based on meta-data?

squirrel/artifact_manager/base.py
SebS94: Main thought here was that people can log this directly to console and view created artifacts in wandb or buckets directly without having to go on a wild goose-chase?

squirrel/artifact_manager/base.py
SebS94: Renamed to `log_artifact` - can be retrieved via `get_artifact`.
In general everything that can be serialised via the backend serialiser (currently messagepack, later deltalake) can be logged. What would be the motivation for restricting this?

SebS94:
> The `default` collection is a bit odd. I assume that the typical usage will be informally assuming some semantic or schema consistency for each collection and organize artifact by that, then default because like `whatever`. I think there is also a risk that people forget to specify the collection and incorrectly put artifacts in default collection, which can be hard to notice and hard to fix.
The collection attribute is intended as an active collection to which objects are logged if no other target is provided. The assumption here is that this is set e.g. to the job id at the beginning. I renamed the attribute to make this more explicit.

> I think a nice extension is for artifact registry would be collections with explicit schema definition, say in `json-schema` format.
Absolutely agree that this will be a useful extension, however, for me this is part of the semantic layer which exposes functionality for defining a schema and retrieving/filtering objects based on it.

> I would suggest to add timestamp as well. We can then add search by time functionality.
This seems useful but will keep it out of the basic logger as it unnecessarily complicates the interface. 
What I think makes sense here is to provide a minimal schema containing basic attributes (e.g. author, timestamp) as part of the semantic layer.

AlirezaSohofi:
> This seems useful but will keep it out of the basic logger as it unnecessarily complicates the interface.
> What I think makes sense here is to provide a minimal schema containing basic attributes (e.g. author, timestamp) as part of the semantic layer.

This should not introduce any change to the interface. The `minimal schema` you mentioned was actually aligned with what I had in mind. Whether it's part of artifact itself or semantic layer is up for discussion and involves trade-offs.

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,100 @@
...
+        The manager maintains an 'active' collections which it logs to by default.
+        To avoid incompatibility between different backends, collections cannot be nested (e.g. as subfolders on a
+        filesystem).
+        """
+        return self._active_collection
+
+    @active_collection.setter
+    def active_collection(self, value: str) -> None:
+        """Do not allow access to anything beyond the root location of the artifact store"""
+        assert re.match(r"^[a-zA-Z0-9\-_:]+$", value), (
```
adrianloy, line 33: minor: Would raise ValueError instead of asserts. Asserts can be turned off

squirrel/artifact_manager/base.py
adrianloy: Could this just be part of the source and be retrieved via `get_artifact_source`? or via string representation of the object?

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,100 @@
...
+        """Upload a collection of file into a (current) collection"""
+
+    @abstractmethod
+    def log_folder(self, files: Path, collection: Optional[str] = None) -> Catalog:
+        """Upload folder as collection of artifacts into store"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def log_artifact(self, obj: Any, name: str, collection: Optional[str] = None) -> Source:
+        """Log an arbitrary python object"""
```
adrianloy, line 87: would be nice if docstring has the serializer we use. Also, are we sure we need this? How are different python versions and pickle versions affecting this? I heard stories that people really dont like pickle for serializing due to version dependencies occurring. Having to figure out which versions were used during logging in able to load a year old artifact seems really painful.

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,100 @@
...
+
+        The manager maintains an 'active' collections which it logs to by default.
+        To avoid incompatibility between different backends, collections cannot be nested (e.g. as subfolders on a
+        filesystem).
+        """
+        return self._active_collection
+
+    @active_collection.setter
+    def active_collection(self, value: str) -> None:
+        """Do not allow access to anything beyond the root location of the artifact store"""
```
adrianloy, line 32: docstring doesnt say much about what the function actually does.

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,100 @@
...
+
+        Maintains a mapping of artifact names to backend objects to facilitate logging and retrieval of arbitrary
+        artifacts.
+        """
+        self._active_collection = "default"
+
+    @property
+    def active_collection(self) -> str:
+        """
+        Collections act as folders of artifacts.
```
adrianloy, line 22: The motivation and intended use-cases of collections could be explained a bit more here, makes it easier for reviewers of this PR to give input even if they were not following the dicussions around this topic.

squirrel/artifact_manager/fs.py
```diff
@@ -0,0 +1,200 @@
...
+    def list_collection_names(self) -> Iterable:
+        """List all collections managed by this ArtifactManager."""
+        return self.backend.keys(nested=False)
+
+    def get_artifact(self, artifact: str, collection: Optional[str] = None, version: Optional[int] = None) -> Any:
+        """Retrieve specific artifact value."""
+        if collection is None:
+            collection = self.active_collection
+        if version is None:
+            version = len(self.backend.complete_key(Path(collection) / Path(artifact)))
```
adrianloy, line 75: I think that might break easily. I would want to be able to delete old versions manually (i.e., do `gsutil rm` ) without then never being able to retrieve the latest version correctly anymore

adrianloy (commented):
Some comments, generally LGTM so far.

SebS94:
@AlirezaSohofi @ThomasWollmann @adrianloy I updated the PR with your comments and also added a basic version of the WandB manager.
Unfortunately, the WandB artifact API is actually somewhat incompatible with what we envisioned. They manage versioned artifacts within collections ("artifact types" in their terminology). Each artifact is actually not atomic but a folder of its own that can contain arbitrary files. 

This goes against what we discussed where each artifact is an individually versioned file or serialised python value. I now somewhat abused the WandB artifact to force each committed value into a separate WandB artifact. While I don't think it's a huge blocker right now, I'd be interested in your input on whether or not we should relax our assumption here and allow multiple files per artifact (not necessarily an issue) or stick with this.

Also let me know if you any additional comments regarding the overall API.

squirrel/artifact_manager/base.py
SebS94: The serialiser depends on the backend - WandB has it's own, for SquirrelFileStores it's actually explicitly chosen when initialising the Manager.

adrianloy:
I would follow their approach. I think wandb integration should be first citizen, as this is likely how we will use it the most. I thought the old artifact manager also coul dbe used for "log everything of this local folder as a single artifact" and I think in general that makes sense. So I think relaxing of our assumptions is fine.

test/test_artifact_manager/wandb_script.py
```diff
@@ -0,0 +1,104 @@
...
+"""
+from pathlib import Path
+
+import tempfile
+import wandb
+from squirrel.artifact_manager.wandb import WandbArtifactManager
+
+src_dir = tempfile.TemporaryDirectory()
+
+test_run = wandb.init(project="sebastian-sandbox")
```
mg515, line 13: Would've been nicer if tests don't require an active wandb run, but to mock it instead.

squirrel/artifact_manager/fs.py
```diff
@@ -0,0 +1,205 @@
...
+        artifact_name: str,
+        local_path: Path,
+        collection: Optional[str] = None,
+        artifact_path: Optional[Path] = None,
+    ) -> Source:
+        """Upload local file or folder to artifact store without serialisation"""
+        if not isinstance(local_path, (str, Path)):
+            raise ValueError("Path to file should be passed as a pathlib.Path object!")
+
+        if artifact_path is not None and not isinstance(artifact_name, (str, Path)):
```
mg515, line 169:
```suggestion
        if artifact_path is not None and not isinstance(artifact_path, (str, Path)):
```

squirrel/artifact_manager/fs.py
```diff
@@ -0,0 +1,205 @@
...
+    def get_artifact(self, artifact: str, collection: Optional[str] = None, version: Optional[str] = None) -> Any:
+        """Retrieve specific artifact value."""
+        raise NotImplementedError(
+            "Logging and retrieving python objects is not yet supported. Please serialize your"
+            "objects and retrieve the logged files with 'download_artifact' instead."
+        )
+
+    def log_files(
+        self,
+        artifact_name: str,
```
mg515, line 160: Surprised to see mandatory `artifact_name` arg. Would it be possible to assign a random name? What are the downsides to it?

test/test_artifact_manager/wandb_script.py
SebS94: I agree - as mentioned at the top of the file this needs to be refactored once I figure out how to properly mock the wandb API. I included it here as a demo of the interface but will remove it in the final commit.

squirrel/artifact_manager/fs.py
SebS94: We can introduce that as well, but when talking to Adrian we decided it doesn't really matter if the user or the artifactmanager generates it - do you want me to alter the behaviour accordingly?

squirrel/artifact_manager/wandb.py
```diff
@@ -0,0 +1,193 @@
...
+    def get_artifact_source(
+        self, artifact: str, collection: Optional[str] = None, version: Optional[str] = None
+    ) -> Source:
+        """Catalog entry for a specific artifact"""
+        if collection is None:
+            collection = self.active_collection
+        if version is None:
+            versions = [
+                instance.version
+                for instance in wandb.Api().artifact_versions(type_name=collection, name=f"{self.project}/{artifact}")
```
mg515, line 75: Getting `Could not find artifact type` here, after invoking `log_files()` and specifying `collection`

full trace
```
  File "/home/miha/squirrel-core/squirrel/artifact_manager/wandb.py", line 75, in get_artifact_source
    for instance in wandb.Api().artifact_versions(type_name=collection, name=f"{self.project}/{artifact}")
  File "/home/miha/chameleon/.venv/lib/python3.9/site-packages/wandb/apis/normalize.py", line 87, in wrapper
    raise CommError(message, err).with_traceback(sys.exc_info()[2])
  File "/home/miha/chameleon/.venv/lib/python3.9/site-packages/wandb/apis/normalize.py", line 41, in wrapper
    return func(*args, **kwargs)
  File "/home/miha/chameleon/.venv/lib/python3.9/site-packages/wandb/apis/public.py", line 936, in artifact_versions
    artifact_type = ArtifactType(self.client, entity, project, type_name)
  File "/home/miha/chameleon/.venv/lib/python3.9/site-packages/wandb/apis/public.py", line 3950, in __init__
    self.load()
  File "/home/miha/chameleon/.venv/lib/python3.9/site-packages/wandb/apis/public.py", line 3984, in load
    raise ValueError("Could not find artifact type %s" % self.type)
```

mg515: I think missing `entity` in `name`, i.e. `name=f"{self.entity}/{self.project}/{artifact}")`

squirrel/artifact_manager/fs.py
mg515: On the second thought, I think it's nicer to choose the name actively and let the version increase take it from there.

squirrel/artifact_manager/wandb.py
```diff
@@ -0,0 +1,193 @@
...
+        files.
+
+        Note: For storing objects it is assumed that wandb.init() has been called so that artifacts can be associated
+        with a run.
+        """
+        super().__init__()
+        if project is not None:
+            self.project = project
+        elif wandb.run is not None:
+            self.project = wandb.run.project
```
mg515, line 29: Some logging would be nice when value is assumed from elsewhere.

squirrel/artifact_manager/wandb.py
```diff
@@ -0,0 +1,193 @@
...
+
+        This assumes that the artifact was logged as a wandb serialised object. If the artifact was a file upload,
+        the file contents should be retrieved using download instead.
+        """
+        raise NotImplementedError(
+            "Logging and retrieving python objects is not yet supported. Please serialize your"
+            "objects and retrieve the logged files with 'download_artifact' instead."
+        )
+
+    def log_files(
```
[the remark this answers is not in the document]
mg515: `local_path` can also be `dir` though, so multiple files get logged. I like the plural form myself better

squirrel/artifact_manager/wandb.py
```diff
@@ -0,0 +1,193 @@
...
+        for artifact in artifact_names:
+            for instance in wandb.Api().artifact_versions(type_name=collection, name=f"{self.project}/{artifact}"):
+                catalog[str(Path(collection, artifact))] = self.get_artifact_source(
+                    artifact, collection, instance.version
+                )
+        return catalog
+
+    def log_artifact(self, obj: Any, name: str, collection: Optional[str] = None) -> Source:
+        """Log serialisable object to artifact store."""
+        raise NotImplementedError(
```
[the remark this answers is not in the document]
mg515: Would've gone for implementation already straight away myself, even if just for serializable built-ins or so. One very common use-case we have is storing datasets/models in cloud storage, and using path as an artifact. This is right now not possible, or am I missing something?

squirrel/artifact_manager/base.py
```diff
@@ -0,0 +1,176 @@
...
+            logger.info(f"Did not log artifact folder at {path} as it seems to be empty.")
+
+        if self.tempdir is not None:
+            self.tempdir.cleanup()
+
+
+class ArtifactManager(ABC):
+    def __init__(self):
+        """Artifact manager interface for various backends."""
+        self._active_collection = "default"
```
mg515, line 66: Does it make sense to expose specifying `active_collection` arg in constructor?

squirrel/artifact_manager/wandb.py
```diff
@@ -0,0 +1,193 @@
...
+            artifact_path = str(artifact_path)
+        if local_path.is_dir():
+            artifact.add_dir(str(local_path), name=artifact_path)
+        else:
+            artifact.add_file(str(local_path), name=artifact_path)
+        artifact.save()
+        artifact.wait()
+        return self.get_artifact_source(artifact_name, collection)
+
+    def download_artifact(
```
mg515, line 170: I thik UX of `download_artifact()` could be improved.

1) Right now it's (by default) downloading artifacts to workdir, which is not ideal. I think context manager with tempdir creation would be preferred. 
2) Additionally, I think we could use `FilePathGenerator` on top, to create iterstream over the downloaded artifacts.

To be honest, I don't fully understand what we are trying to do with `Source` here, seems rather unfriendly to me.

squirrel/artifact_manager/wandb.py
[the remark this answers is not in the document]
mg515: After talking to @SebS94 , decided to leave this for later PRs and implement using references

squirrel/artifact_manager/wandb.py
mg515: After discussion with @SebS94 , decided to implement the context manager and return the artifact path.

Example usage:
```
with self.artifact_manager.download_artifact(artifact=afid, collection=collection) as artdir:
    model = torch.load(Path(artdir) / "state.py", map_location=torch.device(device))
```

squirrel/artifact_manager/base.py
mg515: Sorry for nitpicking, I think the constructor arg should be just `collection`. The `self._active_collection` I think is fine though.

mg515 (approved):
Good to go from my side as a first version, further implementation can go into follow-up PRs. Looking forward to use the new artifact manager woop woop

mg515 (approved):

Issue #143, opened 2023-08-29
[BUG] When forwarding kwargs to `shuffle` the `None` default breaks


jotterbach:
Using squirrel main directly from github.

When using `shuffle` inside a nested function/method, the forwarding of the `shuffle` defaults into the function does not work. According to the doc-strings ```size (int, optional): Buffer size for shuffling. Defaults to 1000. Skip the shuffle step if `size < 2`.``` so I would assume that feeding in `None` would create a default or similar. 

Test file below.

```python
import pytest
from squirrel.iterstream import IterableSource

def get_itersource():
    data = list(range(100))
    source = IterableSource(data)
    return source

@pytest.mark.parametrize('size', [None, 1, 200])
def test_1(size):

    iter = get_itersource().shuffle(size=size).collect()
    assert sorted(iter) == list(range(100))
```

pzdkn:
I'll look into this

pzdkn:
fixed: https://github.com/merantix-momentum/squirrel-core/pull/156

PR #153, opened 2023-10-24
Fix shuffling bug, poetry updates, and upper bound torch version

pzdkn:
# Description
Fix the issue
https://github.com/merantix-momentum/squirrel-core/issues/143

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/squirrel/iterstream/base.py
+++ b/squirrel/iterstream/base.py
@@ -324,11 +324,23 @@ class Composable:
 
         return self.compose(SplitByRank, torch_dist_group)
 
-    def to_torch_iterable(self) -> Composable:
-        """Convert the stream to a torch iterable."""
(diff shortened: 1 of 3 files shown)


squirrel/iterstream/base.py
```diff
@@ -233,8 +233,11 @@ def shuffle(self, size: int, **kw) -> Composable:
         - seed (Union[int, float, str, bytes, bytearray, None]): A data input that can be used for `random.seed()`.
 
         """
-        if size < 2:
+        if size is None:
+            size = 1000
```
AlirezaSohofi, line 237: why not `def shuffle(self, size: Optional[int] = 1000, **kw) -> Composable:` ?

pzdkn: bc of my stupidity 😅

PR #166, opened 2023-12-08
Add multiplexing capabilities to squirrel

jotterbach:
# Description
Addresses issue https://github.com/merantix-momentum/squirrel-core/issues/167

This feature adds a multiplexing functionality that lets the user super and sub-sample datasets. This is necessary if you deal with a mixture of datasets with vastly different sizes.

## Type of change

- [ ] Documentation update

# Checklist:


squirrel/iterstream/multiplexer.py
```diff
@@ -0,0 +1,230 @@
...
+            assert sampling_probas is not None
+            assert len(sampling_probas) == len(
+                composables
+            ), "Need sampling probas and composables to have same number of entries"
+
+        self.composables, self.sampling_probas = self._init_composables_and_probas(
+            sampling_probas, composables, proba_threshold
+        )
+        self.rng = np.random.RandomState(np.random.MT19937(seed=seed))
+        self.buffer: t.List[t.Dict[str, t.Any]] = []
```
AlirezaSohofi, line 102: buffer seems to be unused

squirrel/iterstream/multiplexer.py
```diff
@@ -0,0 +1,230 @@
...
+        self._reinit_counts_ = []
```
AlirezaSohofi, line 103: can we use `Composable.loop()`? note that the object `_LoopIterable` has the attribute `counter` to keep track of the index and `n` for number of re-initialization. `n` being `None` is equivalent to `sys.maxsize`

squirrel/iterstream/multiplexer.py
```diff
@@ -0,0 +1,230 @@
...
+                        mux_it.reinits + 1,
+                    )
+                    logger.info(
+                        f"Reinitializing iterator {mux_it.index} as part of the"
+                        f" {self.mux_strategy} sampling strategy. Now at reinit"
+                        f" count {mux_it.reinits}."
+                    )
+                    if self._iteration_done(_iterators):
+                        logger.info("Multiplexing termination condition reached.")
+                        return
```
AlirezaSohofi, line 230: what about adding a method to describe the current effective ratio of each dataset? we have the index and number of reinits so it should be easy.

squirrel/iterstream/multiplexer.py
```diff
@@ -0,0 +1,230 @@
...
+        2. The maximum oversampling epochs e_i
+
+    Note that these two quanitities are not independent of each other and fixing both
+    conditions at the same time does not guarantee that both are fulfilled at the same
+    time.
+    """
+
+    def __init__(
+        self,
+        composables: t.List[Composable],
```
AlirezaSohofi, line 66: If `MessagepackDriver(url=path).get_iter()` is passed here, each shard is read in each process. There is actually a nice work around, one can pass e.g. `MessagepackDriver(url=path).get_iter(key_hooks=[SplitByWorker]).compose(TorchIterable)` as a composable here (see [this test](https://github.com/merantix-momentum/squirrel-core/blob/86f61c0d403c22c9365e7ca4d37ab179acbac6dd/test/test_driver/test_msgpack.py#L14)). Obviously only a power user would know how to handle it. What do you think about handling this with a simpler api here? An alternative is to document it, but I think it would still requires a lot of effort for the user to learn how to do this properly.

AlirezaSohofi (commented):
Looks really good, thanks a lot for the great contribution!

I added a couple of suggestions, let me know what you think. Some additional thoughts:
- after all composables are iterated over once, we know the true length of each and can adjust the probabilities accordingly. Do you think it's beneficial?
- Current implementation is low-level. The benefit is that it gives the full flexibility to the user, but it also requires a lot of knowledge about the Itersteam and the storage layers to configure everything properly. A prominent example which I added a comment about is spliting by workers. Shuffling on the individual composable is another one. Do you think we should wrap common functionalities? I think every such functionality, if not already possible, should be added to the [get_iter()](https://github.com/merantix-momentum/squirrel-core/blob/86f61c0d403c22c9365e7ca4d37ab179acbac6dd/squirrel/driver/driver.py#L65). We can wrap this and initialize `get_iter` based on the parameters though. I don't have strong opinion about this.

squirrel/iterstream/multiplexer.py
jotterbach: I played around with different options for a bit, but don't see a good way to make this happen.

Following things:

* Using a `.loop()` doesn't let me multiplex as I need to instantiate iterators in a list to choose between them. The `loop` will run until the composable is empty and I cannot selectively just yield from one stream.

* Using an instance of `_LoopIterable` directly will not work as the source will be a `Composable` not an `Iterable`. Hence I loose all of those guarantees.

* As we don't have a container for Composable of Composables I can't use the `loop()` to yield from the underlying composable. That would be the multiplexer that I'm building right now.

So while I like the idea of reusing the `_LoopIterable` I think it's API is not capable enough for this use case without majorly redesigning this PR and the Multiplexer.

squirrel/iterstream/multiplexer.py
jotterbach: Not sure what you mean. If you mean the exact ratios based on the samples, we would need to keep track of the sample counts for each dataset. Without knowing the dataset size and just knowing the number of reinits there is not much we can do. But maybe I'm misunderstanding what you mean?

jotterbach: I added a sample counter from which you then can compute the actual ratio.

jotterbach:
@AlirezaSohofi I addressed your comments and added a few more functionalities. Can you have another look?

squirrel/iterstream/multiplexer.py
AlirezaSohofi: True, the issue I raised is more related to `storage` and `driver` packages not `iterstream`

squirrel/iterstream/multiplexer.py
```diff
@@ -0,0 +1,248 @@
...
+                multiprocessing worker and / or GPU ranks. There are several ways to
+                do this, and we encourage reading the documentation for the corresponding
+                driver or the composable you are using. As an example you can load a
+                message pack drive using:
+                ```python
+
+                c0 = MessagepackDriver(url=local_msgpack_url).get_iter(key_hooks=[SplitByWorker])
+                c1 = MessagepackDriver(url=local_msgpack_url).get_iter(key_hooks=[SplitByWorker])
+
+                mux = Multiplexer([c0, c1]).to_torch_iterable()
```
AlirezaSohofi, line 88:
```suggestion
                mux = Multiplexer([c0, c1]).to_torch_iterable(False, False)
```

AlirezaSohofi: This is necessary to by-pass the check

AlirezaSohofi (approved):
Thanks @jotterbach for the amazing contribution :+1:

squirrel/iterstream/multiplexer.py
AlirezaSohofi: If you are curious why this check even exists: we implemented a similar pipeline for a particular project (not nearly as general as this implementation though) and faced this issue that multiplexing and splitting shard keys at Composable level will make this check to fail

AlirezaSohofi:
@jotterbach a test fails:

```
=================================== FAILURES ===================================
_____________________________ test_sampling_ratio ______________________________
[gw0] linux -- Python 3.9.18 /usr/local/bin/python

dummy_data_catalog = ['data_0', 'data_1', 'data_2']

>   ???

test/test_iterstream/test_multiplexing.py:81: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (['data_0', 'data_1', 'data_2'], 43, 0.0078125, 0.0625), kwargs = {}
initial_draws = 3, start = 1254.793343369, result = None
finish = 1257.402444711, internal_draw_time = 0
runtime = datetime.timedelta(seconds=2, microseconds=609101)
current_deadline = timedelta(milliseconds=2000)

    @proxies(self.test)
    def test(*args, **kwargs):
        self.__test_runtime = None
        initial_draws = len(data.draw_times)
        start = time.perf_counter()
        result = self.test(*args, **kwargs)
        finish = time.perf_counter()
        internal_draw_time = sum(data.draw_times[initial_draws:])
        runtime = datetime.timedelta(
            seconds=finish - start - internal_draw_time
        )
        self.__test_runtime = runtime
        current_deadline = self.settings.deadline
        if not is_final:
            current_deadline = (current_deadline // 4) * 5
        if runtime >= current_deadline:
>           raise DeadlineExceeded(runtime, self.settings.deadline)
E           hypothesis.errors.DeadlineExceeded: Test took 2609.10ms, which exceeds the deadline of 2000.00ms
E           Falsifying example: test_sampling_ratio(
E               dummy_data_catalog=['data_0', 'data_1', 'data_2'],
E               seed=43,
E               p0=0.0078125,
E               p1=0.0625,
E           )

/usr/local/lib/python3.9/site-packages/hypothesis/core.py:804: DeadlineExceeded
---------------------------------- Hypothesis ----------------------------------
WARNING: Hypothesis has spent more than five minutes working to shrink a failing example, and stopped because it is making very slow progress.  When you re-run your tests, shrinking will resume and may take this long before aborting again.
PLEASE REPORT THIS if you can provide a reproducing example, so that we can improve shrinking performance for everyone.

```
At the first glance I don't see why this happens. Any idea?

squirrel/iterstream/multiplexer.py
jotterbach: I already have a fix for that. This current PR is just the preparation for that fix which will come next 👍

jotterbach:
@AlirezaSohofi I increased the timeout deadline for the test. It failed because it probably just took a bad set of random numbers. As it's a statistical test this can happen unfortunately. Hopefully the increased timeout will fix this issue.

AlirezaSohofi (approved):

PR #181, opened 2024-02-15
Ali distributed data

AlirezaSohofi:
# Description

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:


LevUdaltsov:
Please update the description to outline the goal/s of this PR

squirrel/driver/streaming_parquet.py
```diff
@@ -0,0 +1,120 @@
...
+        Note: the `source` argument of scan_parquet is set to self.url + "/*", which assumes non-partitioned
+        parquet
+        """
+        import polars as pl
+
+        return pl.scan_parquet(source=self.url + "/*", **scan_kwargs)
+
+    def query(
+        self,
+        query: str = "SELECT * FROM frame LIMIT 2",
```
mg515, line 77: I would somehow try to handle this easier to debug, i.e. log the query before executing it, especially in the case where the user does not provide `query`, but default is used.

(was looking how method is called in `self.get_iter()` )

squirrel/iterstream/source.py
```diff
@@ -53,6 +54,7 @@ def __init__(
         max_workers: t.Optional[int] = None,
         max_keys: int = 1_000_000,
         max_dirs: int = 10,
+        regex_filter: str = None,
```
mg515, line 57: Would maybe a default value of ("*") make sense? Then you don't need to check for the `regex_filter` if null, but it may slow down iter speed and is possibly more error prone.

squirrel/serialization/parquet.py
```diff
@@ -0,0 +1,26 @@
+
+
+from typing import Any, Iterable, List, Dict
+
+from squirrel.serialization.serializer import SquirrelFileSerializer
+
+
+class ParquetSerializer(SquirrelFileSerializer):
```
mg515, line 8: This is still TODO?

squirrel/store/directory_store.py
```diff
@@ -0,0 +1,49 @@
+import typing as t
+
+from squirrel.serialization.serializer import SquirrelFileSerializer
+from squirrel.store.filesystem import FilesystemStore, get_random_key
+
+
+class DirectoryStore(FilesystemStore):
+    def __init__(self, url: str, serializer: SquirrelFileSerializer, clean: bool = False, **storage_options) -> None:
+        """A Store to access files in a potentially nested directory
```
mg515, line 9: Why is this a store and not just a `Driver`? Since `set()` will always write to only first level, if I get it correctly.

squirrel/store/directory_store.py
```diff
@@ -0,0 +1,49 @@
...
+        if key is None:
+            key = get_random_key()
+
+        fp = f"{self.url}/{key}.{self.serializer().file_extension}"
+
+        if not self._dir_exists:
+            self.fs.makedirs(self.url, exist_ok=True)
+            self._dir_exists = True
+
+        self.serializer().serialize_shard_to_file(obj=value, fp=fp, fs=self.fs)
```
mg515, line 34: More of a broader question, have we ever considered checking if the key exists already before overwriting it? I can imagine it's a little bit of a speed/UX tradeoff.

squirrel/store/directory_store.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def get(self, key: str) -> t.Any:
+        """Retrieve the value of the key using SquirrelSerializer.deserialize_shard_from_file."""
```
mg515, line 37: Not sure how this can access items in a nested dir. Structure is incorporated in `key`?

mg515 (commented):
First round review. I skipped over some of the Parquet and Delta Lake stuff, since I lack some understanding and will leave it for the second round. Also, did not review tests yet.

Not `logging` anything anywhere is a design choice? I think it can often help massively the UX if the user understand what is happening with the underlying program, especially when hitting edge cases.

squirrel/driver/store.py
```diff
@@ -59,6 +61,39 @@ def get_iter(self, flatten: bool = True, **kwargs) -> Composable:
...
+            batch_format="numpy",
+            drop_last=drop_last,
+            batch_size=batch_size,
+            local_shuffle_buffer_size=shuffle_item_buffer,
+            local_shuffle_seed=42,
+        )
+
+        it = IterableSource(ds_iter_batch)
+        if isinstance(self.serializer, PNGSerializer):
+            it = it.map(lambda x: x["image"]).flatten()
```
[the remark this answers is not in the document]
AlirezaSohofi: comes from ray, will try to find a better solution

squirrel/driver/directory.py
```diff
@@ -0,0 +1,56 @@
...
+    """A StoreDriver that by default uses SquirrelStore with messagepack serialization."""
+
+    name = "directory"
+
+    SERIALIZERS = {"npy": NumpySerializer(), "png": PNGSerializer()}
+
+    def __init__(self, url: str, file_format: str, storage_options: dict[str, Any] | None = None, **kwargs):
+        """
+        A driver to access a directory of files in the same file formats. the class attribute
+        SERIALIZE contains the supported formats along with the corresponding serializerd from
```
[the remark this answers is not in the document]
AlirezaSohofi: This is starting to be a it dirty now, previously serializer, store, and driver had one-to-one mapping, and file extensions where always the same (hardcoded `.gz`) but now we need to refactor it. Happy to get suggestions about this.

squirrel/driver/store.py
```diff
@@ -95,3 +130,38 @@ def store(self) -> AbstractStore:
...
+
+    def get_ray_dataset(self) -> Dataset:
+        """Get a ray dataset"""
+        import ray
+
+        ds = ray.data
+        if isinstance(self.serializer, PNGSerializer):
+            ds = ds.read_images(self.url)
+        elif isinstance(self.serializer, NumpySerializer):
+            ds = ds.read_numpy(self.url)
```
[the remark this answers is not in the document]
AlirezaSohofi: it definitely needs some refactoring and cleaning up :+1:

squirrel/iterstream/source.py
AlirezaSohofi: I didn't do that exactly because of speed.

squirrel/serialization/parquet.py
AlirezaSohofi: no it's a by-product of `new things not quite fitting the old abstractions`. For json and msgpack we need to serialize data, but parquet takes care of it automatically and we don't want to touch that, but a driver and store expect to have `Serializer`. Needs refactoring and happy to get suggestions

squirrel/store/directory_store.py
AlirezaSohofi: No we don't. In most cases you leave the `key` argument empty and it will be randomly generated. If a user do wants specific key she should take care of not overwriting existing ones. Out of scope for Squirrel I'd say.

squirrel/store/directory_store.py
AlirezaSohofi: yes, it's `def keys(..., nested = True)`. I think `nested` is a bit confusing in general. We want to make it easy to works with nested directories because stores being extended over time are safer if each time they write the new data into a dedicated sub directory, but it can get confusing. It's a broader concern though. I can move the `Directory` into a separate pr for more discussion if you find it useful.

mg515: I think `nested` is fine here tbh. I get a bit confused over which use case this is supposed to cover. Where one wants to write subsets of dir (maybe train, val splits, or some other division) and then iter only specific keys based on regex?

Check also other comment above - while reading is somewhat enabled, the writing of substores is not, only maybe through specifying custom keys - but even then I think we right now don't create dirs if they don't exist. Maybe test such a case more thoroughly.

squirrel/driver/store.py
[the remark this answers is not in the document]
mg515: at minimum put them under `constants.py` or so

squirrel/driver/directory.py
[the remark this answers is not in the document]
mg515: As for the `Serializer` - I think it's fine to keep the name and be able to go both directions.

As of one-to-one mapping, I don't necessarily see the issue with one-to-many for stores, which I think was already the case before (e.g. `SquirrelStore` can work with msgpack or jsonl serializer). As for drivers, I'm not sure, as I rarely used them outside of store context.

squirrel/driver/directory.py
```diff
@@ -0,0 +1,56 @@
...
+    "DirectoryDriver",
+]
+
+
+class DirectoryDriver(StoreDriver):
+    """A StoreDriver that by default uses SquirrelStore with messagepack serialization."""
+
+    name = "directory"
+
+    SERIALIZERS = {"npy": NumpySerializer(), "png": PNGSerializer()}
```
mg515, line 22: I think we should avoid using `PNG`, but just make it a more general `ImageSerializer` and use `skimage.io.imread` underneath (which we already do), and allow all common image extensions.

squirrel/driver/directory.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+        if "store" in kwargs:
+            raise ValueError("Store of MessagepackDriver is fixed, `store` cannot be provided.")
+
+        self.serializer = self.SERIALIZERS.get(file_format, None)
+        if self.serializer is None:
+            raise ValueError(
+                f"""the file_format argument {file_format} is invalid,
+                valid file_formats are {''.join(list(self._SERIALIZERS.keys))}"""
+            )
```
mg515, line 38:
```suggestion
                valid file_formats are {''.join(list(self.SERIALIZERS.keys))}"""
```
'DirectoryDriver' object has no attribute '_SERIALIZERS'

squirrel/store/directory_store.py
```diff
@@ -0,0 +1,49 @@
...
+        fp = f"{self.url}/{key}.{self.serializer().file_extension}"
```
mg515, line 38:
```suggestion
        fp = f"{self.url}/{key}.{self.serializer.file_extension}"
```

squirrel/store/directory_store.py
```diff
@@ -0,0 +1,49 @@
...
+        return self.serializer().deserialize_shard_from_file(fp=fp, fs=self.fs)  # f"{self.url}/{key}"
```
mg515, line 39:
```suggestion
        fp = f"{self.url}/{key}.{self.serializer.file_extension}"
        return self.serializer.deserialize_shard_from_file(fp=fp, fs=self.fs)  # f"{self.url}/{key}"
```

squirrel/driver/store.py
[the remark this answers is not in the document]
LevUdaltsov: Agreed we should avoid hardcoding, using enum here would be better

test/conftest.py
```diff
@@ -250,3 +257,112 @@ def dummy_data_catalog(tmp_path_factory: TempPathFactory) -> Catalog:
...
+
+
+@pytest.fixture
+def normal_parquet_ray() -> Iterable:
+    """A parquet dataset without np.array written using Ray"""
+    _data = get_records_without_np(NUM_ROWS)
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        ds = ray.data.from_items(_data)
+        ds.write_parquet(tmp_dir)
+        yield tmp_dir, _data
```
LevUdaltsov, line 368: I had to `pip install grpcio` for this to work, should maybe add to dep list. unsure why ray doesnt install it, did you have any issues here?

test/test_driver/test_parquet_driver.py
LevUdaltsov, line 1: nice tests, running through these with debugger made it clearer how these things are supposed to work together

LevUdaltsov:
Nice work, cool to see how easy we can extend squirrel with these other frameworks. as a POC it is a great point of reference.

I think this PR is adding too much at once and I am myself not clear which of these tools we wish to support and why. For example is there a use case for us to use `Polars` or `Deltalake` already? I would rather add Ray first, and then add the other two separately. I had not used either until poking around this PR. Adding them one at a time would make it easier to focus on one change and then to communicate what it does. 

Even with Ray itself, it is not clear to me whether handling `Iterables` created by ray should be done by `StoreDriver` or somewhere else, in another class. Figuring out what we want to achieve by including it is more important than how we do it - this is not clear yet to me. 

I would be in favour of a brainstorm on this and then perhaps a refactor of how the drivers and serializers are constructed, and then plugging Ray and other frameworks into that. Input from how Ray has been used for data loading in other projects would be helpful here. WDYT @mg515 @AlirezaSohofi

squirrel/driver/streaming_parquet.py
```diff
@@ -0,0 +1,120 @@
...
+    name = "streaming_parquet"
+
+    def __init__(self, url: str, storage_options: Optional[Dict[str, Any]] = None, **kwargs):
+        """Driver to stream data from a parquet dataset."""
+        super().__init__(url=url, serializer=ParquetSerializer(), storage_options=storage_options, **kwargs)
+
+    @property
+    def store(self) -> ParquetStore:
+        """Return squirrel.store.ParquetStore"""
+        return ParquetStore(self.url, **self.storage_options)
```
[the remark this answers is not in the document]
AlirezaSohofi: The reason for not initializing store in `__init__` is that this object must be serializable, otherwise it can't be used in pytorch Dataloader in multiprocessing mode, i.e. the following will fail

```
it = StreamingParquetDriver(url).get_iter().to_pytorch_dataloader()
dl = tut.Dataloader(it, num_workers=2)  # if you omit num_worker it works but will be single process
```

squirrel/driver/streaming_parquet.py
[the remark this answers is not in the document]
AlirezaSohofi: No strong opinion, slight preference for more explicit naming. If we opt for a uniform naming would call it something other than `load_data`, as it could be misleading because it doesn't really load, but just returns the object to interact with the underlying storage

squirrel/serialization/serializer.py
```diff
@@ -7,17 +7,28 @@
...
+
+
+class SquirrelSerializer(SquirrelFileSerializer):
+    @abc.abstractmethod
+    def serialize(self, obj: t.Any) -> t.Any:
+        """Returns the serialized object."""
+
+    @abc.abstractmethod
+    def deserialize(self, obj: t.Any) -> t.Any:
+        """Returns the deserialized object."""
```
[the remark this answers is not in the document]
AlirezaSohofi: 1. different stores need some but not all of the methods of `Serializer`, so the alternative is to raise `UnimplementedError` in some places
2. good suggestion. Let's do that in the follow up refactoring pr

squirrel/store/parquet_store.py
```diff
@@ -0,0 +1,96 @@
...
+from squirrel.iterstream.source import FilePathGenerator
+from squirrel.store.filesystem import FilesystemStore
+
+
+class ParquetStore(FilesystemStore):
+    def __init__(self, url: str, **storage_options):
+        """Store that uses pyarrow to read from / write to the dataset"""
+        super().__init__(url=url, serializer=None, storage_options=storage_options)
+
+    def get(self, key: str, dataset_kwargs: t.Optional[t.Dict] = None, **open_kwargs) -> t.Any:
```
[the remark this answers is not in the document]
AlirezaSohofi: True, will fix

squirrel/store/parquet_store.py
```diff
@@ -0,0 +1,96 @@
...
+            mode (str): passed to write_deltalake(), one of the 'error', 'append',
+                'overwrite', 'ignore', default to append
+            kwargs: passed to write_deltalake
+        """
+        import pyarrow as pa
+        from deltalake import write_deltalake
+
+        rec_batch = pa.RecordBatch.from_pylist(value)
+        # TODO: can we use get_fs_from_url(self.url)?
+        return write_deltalake(f"{self.url}", rec_batch, mode=mode, **kwargs)
```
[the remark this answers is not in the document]
AlirezaSohofi: get is the same as the parent class

test/conftest.py
AlirezaSohofi: Strange! I don't think I had this issue

test/test_driver/test_parquet_driver.py
```diff
@@ -0,0 +1,208 @@
...
+    """Assert arrays are equal"""
+    _arrays = [np.array(sorted(list_of_arrays, key=lambda x: np.sum(x))) for list_of_arrays in arrays]
+    # print(f"...................{len(_arrays)} {', '.join([str(_a.shape) for _a in _arrays])}")
+    assert np.sum(_arrays - _arrays[0]) == 0
+
+
+def test_deltalake_store(test_path: URL) -> None:
+    """Test deltalake store"""
+    _data = get_records_without_np(10)
+    dst = DeltalakeStore(test_path)
```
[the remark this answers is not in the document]
AlirezaSohofi: no its a bug, thanks for catching it. Will fix it :+1:

squirrel/store/parquet_store.py
```diff
@@ -0,0 +1,96 @@
...
+    def set(self, value: t.Any, **open_kwargs) -> None:
+        """Persists an item with the given key.
+
+        Args:
+            value (Any): Item to be persisted.
+            dataset_kwargs: Keyword arguments that will be forwarded to the `pyarrow.parquet.write_to_dataset`.
+            **open_kwargs: Keyword arguments that will be forwarded to the filesystem object when opening the file.
+        """
+        import pyarrow as pa
+        import pyarrow.parquet as pq
```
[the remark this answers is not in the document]
LevUdaltsov: oh nice, thanks for the pointer. Lets leave as is then

squirrel/serialization/serializer.py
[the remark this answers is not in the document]
LevUdaltsov: What stores would need only a subset?

squirrel/driver/streaming_parquet.py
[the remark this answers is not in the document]
LevUdaltsov: Depends on use case I guess. Should users know specifically what driver they are interact with before calling the method?

PR #196, opened 2024-05-07
Bump python and pandas version

MaxSchambach:
# Description

Currently, pandas is fixed as `pandas = "<=1.5.3"` which prevents using squirrel with other more up-to-date libraries. The reason for this pin, according to the current specification, is "pin for dask compatibility (afterwards core.strings.Stringmethods is not available anymore)", see [here](https://github.com/merantix-momentum/squirrel-core/blob/v0.19.8/pyproject.toml#L32).

This has been resolved in newer Dask versions, see [here](https://github.com/dask/dask/issues/10164) and [here](https://github.com/eurec4a/how_to_eurec4a/issues/92).

Hence, pandas and dask versions are bumped. Python version is bumped to 3.9 min, due to pandas requirement.


LevUdaltsov (approved):
LGTM!

PR #185, opened 2024-03-07
directorydriver

AlirezaSohofi:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. 
List any dependencies that are required for this change.

Fixes # issue

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring including code style reformatting 
- [ ] Other (please describe):

# Checklist:

- [ ] I have read the [contributing guideline doc](https://squirrel-core.readthedocs.io/en/latest/developer/code_of_conduct.html) (external contributors only)
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have kept the PR small so that it can be easily reviewed  
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] All dependency changes have been reflected in the pip requirement files.

PR #199, opened 2024-05-16
Feat/ray dataset from get iter

MsMatias:
# Description

This pull request introduces a new capability to the get_iter method by implementing support for the Ray framework. With this implementation, the method now provides flexibility in choosing between two different engines: Ray and iterstream. Users can specify the engine parameter as either 'ray' or 'iterstream' to select the desired processing framework. When 'ray' is chosen, the method returns a ray.data.Dataset, enabling parallelized processing of data items. This enhancement enhances the scalability and performance of data processing tasks, particularly when dealing with large datasets. Additionally, the implementation ensures backward compatibility by retaining support for the iterstream engine.

Changes made by @ThomasWollmann 

Fixes # issue

## Type of change

- [ ] Documentation update

# Checklist:

### DIFF
--- a/squirrel/driver/driver.py
+++ b/squirrel/driver/driver.py
@@ -2,14 +2,13 @@
 
 from __future__ import annotations
 
+import ray
+from ray.data import Dataset
 import functools
 from abc import ABC, abstractmethod
-from functools import partial
-from inspect import isclass
 from typing import Any, Callable, Iterable
 
 from squirrel.catalog import Catalog
-from squirrel.iterstream import Composable, IterableSource
 
 
 __all__ = ["Driver", "IterDriver", "MapDriver"]
@@ -32,8 +31,8 @@ class IterDriver(Driver):
     """
 
     @abstractmethod
-    def get_iter(self, **kwargs) -> Composable:
-        """Returns an iterable of items in the form of a :py:class:`Composable`, which allows various stream
+    def get_iter(self, **kwargs) -> Dataset:
+        """Returns an iterable of items in the form of a :py:class:`Dataset`, which allows various stream
         manipulation functionalities.
 
         The order of the items in the iterable may or may not be randomized, depending on the implementation and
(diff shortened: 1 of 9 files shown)


MaxSchambach (commented):
Thank you for the contribution! I am not too deep into the low-level workings of squirrel, so take this with a grain of salt:

To me, this seems to be a pretty drastic change to the API (even though it is backward-compatible) and I don't fully understand the necessity quite yet. 
The downsides of this approach are:

 - the API becomes inconsistent in my opinion, as `get_iter` now has two distinct return types. This makes the subsequent use of the iterable dependent on the engine used, e.g. batching, splitting, (async) maps, etc. This is a drastic change that is likely not expected by a user for which changing the engine is rather subtle.
 -  `ray` effectively becomes a non-optional dependency, which is bloating `squirrel` for users not requiring `ray`. This is undesireable in my opinion.

Why is the ray integration not achieved via custom drivers implementing a custom `get_iter` method instead? This is similar to the Huggingface or Torchvision drivers in [squirrel-datasets-core](https://github.com/merantix-momentum/squirrel-datasets-core/tree/main/src/squirrel_datasets_core/driver)? I understand, that the application here is slightly different, but I feel like having a driver around ray Datasets instead of changing the `get_iter` engine could resolve the above shortcomings. Also, it would be nice to still maintain the Compasable return type as to not having to change the subsequent processing pipelines.

squirrel/driver/driver.py
```diff
@@ -45,7 +47,7 @@ class MapDriver(IterDriver):
     """A Driver that allows retrieval of items using keys, in addition to allowing iteration over the items."""
 
     @abstractmethod
-    def get(self, key: Any, **kwargs) -> Any:
+    def get(self, key: Any, **kwargs) -> BaseModel:
```
MaxSchambach, line 50: I don't understand the need of pydantic here.

MsMatias: We can ensure that the returned objects are inherited from the basemodel type. Just to make it more restrictive. But We can stripped it away.

MaxSchambach: I still don't understand 🙃
The return values of the `get` method of inherited classes will not be `BaseModel` (nor classes inherited from `BaseModel`), but, mostly, some sort of iterator, see e.g. the `StoreDriver` or `ZarrDriver` implementations [here](https://github.com/merantix-momentum/squirrel-core/blob/main/squirrel/driver/store.py#L62) and [here](https://github.com/merantix-momentum/squirrel-core/blob/main/squirrel/driver/zarr.py#L76). Hence, I believe the `Any` type annotation is less confusing. And, as far as I know, `pydantic` is not yet a dependency of squirrel.

AlirezaSohofi (commented):
Thank you so much for the nice implementation. 

There is another attempt at adding `ray` support for squirrel #181 . The main difference is here you extended `get_iter()`, but in the other pr I decided for [get_iter_ray](https://github.com/merantix-momentum/squirrel-core/pull/181/files#diff-c819b755771d93e27d1053cab4ed5996e1c977fe2ccc8b8616f331e2f70a1cc8R101). The motivations are:

- iterstream and ray provide similar functionalities such as `shuffle` and `prefetch`, but they have different semantics. I could not fit the arguments of `iter_stream` into ray and thought it might be confusing or even harmful.
- Similarly ray has some functionalities and arguments that iterstream does not support. For instance one can a[pply a function on a batch of data](https://github.com/merantix-momentum/squirrel-core/pull/181/files#diff-c819b755771d93e27d1053cab4ed5996e1c977fe2ccc8b8616f331e2f70a1cc8R137) which can boost performance considerable.
- Apart from clarity and performance, there is another potentially critical point about memory utilization. Squirrel encourages big shards, several hundred MG is fairly common. It's no problem for iterstream because it does not load this into memory, but streams from the file directly. See [here](https://github.com/merantix-momentum/squirrel-core/blob/main/squirrel/store/squirrel_store.py#L41) for example. With ray it's not really possible to do that. You have to load the entire shard into memory all at once, as you did [here](https://github.com/merantix-momentum/squirrel-core/pull/199/files#diff-3ebed2c94924925c2a35d747ba4810e7054e19c4af9398fa790bf09e526334d3R171) too. This is designed so because it avoid GPU cold start, significantly improves IO when reading from s3/gcs, and provides better compression that lowers storage. It would be awesome if you have any idea how to reconcile these two modes.

I think there are two main question we need to answer:
- what it the data source? if it's many small files your implementation makes total sense, if small number of big shards, maybe not. 
- how the stream is going to be consumed? If it's meant to be used with pytorch, I don't expect the current implementation to provide big speedup, as there is an overhead of serialization/deserialization, possible global data movement, and more network communication (assuming distributed ray). If it's going to be used with ray train, which is imo super interesting to explore, then current squirrel abstractions do not contribute much, it's probably better to use pure ray than wrapping it in a very thin layer that doesn't do much. 

This [pr](https://github.com/merantix-momentum/squirrel-core/pull/185/files) might also be interesting.

AlirezaSohofi:
@MaxSchambach 
> Why is the ray integration not achieved via custom drivers implementing a custom `get_iter` method instead?

I think ray is equivalent to `iterstream`, so if we want to support it, it has to be part of squirrel itself. It can still be an optional dependency.

> the API becomes inconsistent in my opinion, as get_iter now has two distinct return types. 

Totally agree with this point.

MaxSchambach:
@AlirezaSohofi 
> I think ray is equivalent to `iterstream`, so if we want to support it, it has to be part of squirrel itself. It can still be an optional dependency.

Makes sense! In that case, I would second your approach and introduce a `get_ray_iter` or `get_ray_dataset` method instead.

adrianloy:
I think it was up for discussion to drop the iterstream functionality completely. Maybe it makes sense to be aligned on that before discussing implementation details here.

squirrel/driver/driver.py
```diff
@@ -120,44 +123,87 @@ def get_iter(
...
+
+
+        if engine == 'ray':
+            # read keys in a generator to avoid memory overflow
+            def _key_read_f(x):
+                for y in self.store.keys():
+                    yield {'item': y}
+            it = ray_range(1).flat_map(_key_read_f)
+            num_keys = it.count()
+            it = it.random_shuffle()
```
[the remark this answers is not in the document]
mg515: Using `.random_shuffle()` is also the least efficient way to shuffle, see why in [ray docs](https://docs.ray.io/en/latest/data/shuffling-data.html#shuffle-all-rows).

Shuffling should work either by shuffling the keys (random order of reading the shards / rows), and/or shuffling the items within shards. This can be achieved by specifying `local_shuffle_buffer_size` argument at `iter_batches()` within ray and similarly for squirrel as `iter_kwargs`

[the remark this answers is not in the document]
MsMatias: Thank you all! 

@AlirezaSohofi I'll add the parameters due to it would help the developer to have a way to control the shuffling. Good idea!

@mg515 I see... fair point! Something I'd like to add and based on what @AlirezaSohofi mentioned, we can add an argument to let the user select what kind of shuffling they want, whether the turn on the shuffling or not then they will need to call it later by `iter_batches()`. My argument why we shouldn't remove it, it's because the global shuffling is way too different from the local shuffling, and I'm not sure if the `iter_batches()` can do global shuffling.

PR #222, opened 2025-01-23
Fix deprecated wandb api call

mg515:
```
wandb: WARNING Api.artifact_versions(type_name, name) is deprecated, use Api.artifacts(type_name, name) instead.
```


mg515:
Should be exactly the drop-in replacement

```

    @normalize_exceptions
    def artifact_versions(self, type_name, name, per_page=50):
        """Deprecated, use artifacts(type_name, name) instead."""
        deprecate(
            field_name=Deprecated.api__artifact_versions,
            warning_message=(
                "Api.artifact_versions(type_name, name) is deprecated, "
                "use Api.artifacts(type_name, name) instead."
            ),
        )
        return self.artifacts(type_name, name, per_page=per_page)

    @normalize_exceptions
    def artifacts(self, type_name, name, per_page=50):
        entity, project, collection_name = self._parse_artifact_path(name)
        return public.Artifacts(
            self.client, entity, project, collection_name, type_name, per_page=per_page
        )
```

MsMatias (approved):
lgtm. not so much I can contribute yet

